PR #81, opened 2026-04-30
fix(ci): unblock pre-existing armv6m + rp2040-pio CI failures

w1ne:
## Summary

Two pre-existing main-branch CI failures that block every PR. Both are **unrelated** to recent ESP32-S3 work — they surface now because the affected jobs are path-filtered to PR builds, and main never re-ran them after the regressions landed.

### 1. `coverage-matrix-ci-fixture-armv6m`

The workspace `.cargo/config.toml` unconditionally passes `-C link-arg=-Tlink.x` for every `thumbv6m-none-eabi` build (added Mar 10 in `7197647` for cortex-m-rt firmwares). `firmware-ci-fixture` targets thumbv6m but doesn't depend on cortex-m-rt — so `link.x` doesn't exist for it, and rust-lld errors out:

```
rust-lld: error: cannot find linker script link.x
```

**Fix:** have `firmware-ci-fixture/build.rs` emit an empty `link.x` stub. The real memory layout still comes from `-Tminimal.ld` already passed by build.rs.

### 2. `onboarding-rp2040-pio` (UART captures 0 bytes → `PIO_OK` never seen)

`firmware-rp2040-pio-onboarding` wrote UART data to offset `0x00` (correct PL011/RP2040 hardware), but `configs/chips/rp2040.yaml` models UART0 with `profile: "stm32v2"` — a deliberate simulator shortcut whose TDR register is at offset `0x28`. Writes to `0x00` fell into the CR1 byte-tracking path of the STM32-V2 model and produced no UART output.

**Fix:** align the firmware with the simulator convention (TDR at `0x28`), matching its sister `firmware-rp2040-demo`. Aliasing offset `0x00` → TX in the UART model would break stm32h563/h5 firmwares that legitimately write CR1 there.

## Test plan

- [x] `cargo build -p firmware-ci-fixture --release --target thumbv6m-none-eabi` → succeeds
- [x] `bash scripts/onboarding_smoke.sh ... --target-id rp2040-pio ...` → status: pass, UART captured `PIO_OK`
- [x] `cargo test -p labwired-core --test firmware_survival test_rp2040_demo_survival --release` → still passes (regression check for the sister rp2040 firmware that already used offset 0x28)
- [x] Workspace `cargo test --workspace --release` → 3 pre-existing failures (`test_spi_fidelity_in_machine`, `test_strict_board_onboarding`, `test_register_compliance_all_chips`) all also fail on plain `90e09aa` (current main HEAD), so they're not regressions from this PR

## Out of scope

- The two pre-existing test failures above (`test_spi_fidelity_in_machine`, `test_strict_board_onboarding`) — separate cleanup
- Adding a proper PL011 UART model for RP2040 (the chip yaml's `stm32v2` profile is a known simulator shortcut)

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,0 @@
-# Linker configuration for ARM targets.
-# cortex-m-rt 0.7.x does not emit rustc-link-arg=-Tlink.x automatically;
-# this config.toml provides it globally for all thumbv6m targets.
-
-[target.thumbv6m-none-eabi]
-rustflags = ["-C", "link-arg=-Tlink.x"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,6 @@
+# Linker configuration for ARM targets.
+# cortex-m-rt 0.7.x does not emit rustc-link-arg=-Tlink.x automatically;
+# this config.toml provides it globally for all thumbv6m targets.
+
+[target.thumbv6m-none-eabi]
+rustflags = ["-C", "link-arg=-Tlink.x"]
--- a/crates/firmware-ci-fixture/build.rs
+++ b/crates/firmware-ci-fixture/build.rs
@@ -21,6 +21,10 @@ fn main() {
         .unwrap()
         .write_all(include_bytes!("minimal.ld"))
         .unwrap();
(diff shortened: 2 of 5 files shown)


w1ne:
Superseded by PR #83 (which folds in this PR's two commits as part of a comprehensive 'unblock CI' patch). Closing.

PR #94, opened 2026-05-23
sim: portYIELD + timer tick + abort-halt — fix IPC-task spin (DC7)

w1ne:
## Summary
- Implement `WSR.INTSET` (id 226) — was treated as ignored direct-INTERRUPT write, dropping every `portYIELD()` silently
- CCOMPARE0 ack-and-rearm on write — fixes the timer tick when firmware sets the comparator below the current CCOUNT (bump-allocator boot overruns the divisor)
- New `esp_clk_cpu_freq` thunk returning 240 MHz — without it `_xt_tick_divisor` is 0 and the tick ISR re-fires every cycle
- New `abort_halt` thunk for `__assert_func` / `abort` / `panic_abort` — these are no-return on real silicon; stubbing as nop turned `xQueueGenericSend`'s NULL-queue assertion into a tight retry loop
- Stub `xQueueGiveMutexRecursive` / `xQueueTakeMutexRecursive` / `esp_log_impl_*` — newlib stdio locks and the log mutex back onto recursive mutexes whose handles are never initialised in the bump-allocator world
- Drop the dead inherent `SystemBus::write_u32/write_u16` watchpoint from #93 — it never fired because the CPU dispatches through the `Bus` trait impl, not the inherent method

## Background
DC7 was diagnosed as an SMP race in vListInsert. The actual mechanism: an IPC task (single-core, on PRO_CPU) calls `xQueueReceive` on an empty queue, `vTaskPlaceOnEventList` adds it to the wait list, exits the critical section — and then re-enters because the yield was silently dropped. The wait list ends up containing the same item twice; the walker walks into the item itself, and the second insertion writes `item.pxNext = item` (self-reference). Subsequent inserts spin forever in the walker.

With portYIELD now working, timer ticks firing on schedule, and assertions actually halting instead of looping, the spin pattern is broken. The next visible blocker is unrelated firmware-init details (display library init paths) hitting fresh real assertions, which now produce clean halts with diagnostics rather than infinite loops.

## Test plan
- [x] `cargo test --workspace --release --tests` — 913 tests pass
- [x] Reader sketch sim with these fixes: PC progresses out of the vListInsert spin into setup() and onward (then halts at a downstream firmware assertion with a clear diagnostic)
- [ ] AgentDeck snapshot path still works end-to-end in the playground (untested in this PR — please verify before deploying)

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1071,16 +1071,26 @@ fn run_snapshot_capture(args: SnapshotCaptureArgs) -> ExitCode {
     } else if args.profile == "agentdeck" {
         thunks.push((0x400e_2280, rom_thunks::nop_return_zero));
     }
-    // ESP-IDF clock/efuse/cache/dport bring-up — the sim has no silicon
-    // behind these so we stub them to return-0. Only installed when the
-    // symbol is present in the ELF (Arduino-ESP32 profile).
+    // Real-silicon noreturn functions — abort_halt prints diagnostics and
+    // halts the CPU instead of returning. Without this, stubbing them as
+    // nop_return_zero creates tight `assert → return → re-check → assert`
+    // loops in xQueueGenericSend's parameter-validation path.
     for sym in &[
         "panic_abort",
-        "__assert_func",   // newlib: catches double-fault on assert during reent init
+        "__assert_func",
         "abort",
         "__assert",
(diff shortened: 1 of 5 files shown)


w1ne:
Superseded by #96 (folded into the v0.15.0 release branch). Commit `fc8feed` is now on main.

PR #95, opened 2026-05-23
sim: get past FreeRTOS startup into Arduino user code

w1ne:
## Summary
- Implement `WSR.INTSET` (id 226) — was treated as ignored direct-INTERRUPT write, dropping every `portYIELD()` silently
- CCOMPARE0 ack-and-rearm on write — fixes the timer tick when firmware sets the comparator below the current CCOUNT
- New `esp_clk_cpu_freq` thunk returning 240 MHz so `_xt_tick_divisor` is non-zero
- `abort_halt` thunk for `__assert_func` / `abort` / `panic_abort` — clean halt instead of nop-return + tight assert loop
- `xQueueCreateMutexStatic` echoes its static-buffer arg back; `xTaskGetCurrentTaskHandle` reads `pxCurrentTCB[core]`; `xQueueSemaphoreTake` / `xQueueGenericSend` return pdTRUE on NULL queues; `esp_ipc_init` skipped; recursive-mutex + esp_log_impl wrappers stubbed
- HardwareSerial / UART layer stubbed (Serial.print no-ops at the leaf, doesn't break Print's virtual dispatch)
- SPIClass::beginTransaction thunk lazily inits `this->_spi` to a fake spi_t with dev=SPI3 base AND programs USER_REG bit 27 (USR_MOSI) so the spi3 peripheral actually fires bytes to attached devices

## Result
Reader sketch boots through FreeRTOS, reaches `setup()`, runs `drawPage()`, drives the SSD1680 panel via SPI, and renders visible text glyphs:

| metric              | baseline | after          |
|---------------------|----------|----------------|
| reader spi3 txns    | 0        | **19039**      |
| reader refreshes    | 0        | **3 cycles**   |
| reader non-FF bytes | 0        | **1429/4736**  |
| AgentDeck spi3 txns | 19031    | 19031          |
| AgentDeck refreshes | 1        | 1              |
| AgentDeck non-FF    | 782      | 782            |
| core tests          | 347 ✅   | 347 ✅          |

## Background
Builds on #94's portYIELD / timer / abort_halt foundation. With clean halts on assertions instead of infinite loops, this round walked one assertion at a time. The last fix needed care — the initial round stubbed Print/Stream wholesale to get past Serial.println spin, but that broke `display.print()` too because `Print::print → Print::write → Adafruit_GFX::write → drawChar → drawPixel` all rides the same dispatch chain. Final fix: only stub the HardwareSerial / uart layer at the leaves; Print and Stream stay real.

## Regression check
AgentDeck snapshot path identical to baseline (spi3=19031, refresh=1, 782 non-FF bytes). All new thunks gate on auto-discovered symbols so AgentDeck's stripped ELF is untouched.

## Test plan
- [x] `cargo test -p labwired-core --release` — 347 tests pass
- [x] Reader sketch: reaches `drawPage()`, 19039 SPI txns, 3 refreshes, 1429 non-FF bytes
- [x] AgentDeck regression — identical metrics to baseline

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1071,16 +1071,26 @@ fn run_snapshot_capture(args: SnapshotCaptureArgs) -> ExitCode {
     } else if args.profile == "agentdeck" {
         thunks.push((0x400e_2280, rom_thunks::nop_return_zero));
     }
-    // ESP-IDF clock/efuse/cache/dport bring-up — the sim has no silicon
-    // behind these so we stub them to return-0. Only installed when the
-    // symbol is present in the ELF (Arduino-ESP32 profile).
+    // Real-silicon noreturn functions — abort_halt prints diagnostics and
+    // halts the CPU instead of returning. Without this, stubbing them as
+    // nop_return_zero creates tight `assert → return → re-check → assert`
+    // loops in xQueueGenericSend's parameter-validation path.
     for sym in &[
         "panic_abort",
-        "__assert_func",   // newlib: catches double-fault on assert during reent init
+        "__assert_func",
         "abort",
         "__assert",
         "__cxa_pure_virtual",
         "__cxa_throw",
+    ] {
+        if let Some(&pc) = symbol_addrs.get(*sym) {
+            thunks.push((pc, rom_thunks::abort_halt));
+        }
+    }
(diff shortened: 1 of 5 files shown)


w1ne:
Superseded by #96 (folded into the v0.15.0 release branch). Commits `412cbcd`, `1e73174`, `c4b7fb8` are now on main.

PR #142, opened 2026-05-30
perf: cache rtc_cntl index + word-granular Dport/Timg/RtcCntl + dense DPORT storage

w1ne:
## Summary

Profiler on the ereader bench showed the Rust CPU interpreter was only ~4.5% of wall time; the rest was orchestration overhead. Three small, surgical fixes here close most of that gap:

1. **Cache RTC_CNTL peripheral index in `Machine`** (`crates/core/src/lib.rs`).
   `drain_rtc_cntl_reset_request` used to walk the full ~40-peripheral list every CPU cycle, downcasting each to `RtcCntl` to check a latch that's almost never set. The index is now resolved once at `Machine::new`; the per-cycle path becomes one indexed access + one downcast (and short-circuits to `false` for non-ESP32-classic configs with no RtcCntl on the bus).

2. **Word-granular `read_u32` / `write_u32` overrides on `Dport`, `Timg`, `RtcCntl`** (`crates/core/src/peripherals/esp32/{dport,timg,rtc_cntl}.rs`).
   The default `Peripheral::read_u32`/`write_u32` issues four byte ops, each hashing through SipHash. The bench polls DPORT and TIMG word registers every cycle, so that's 16 SipHash hashes per polled word. Overrides collapse this to one lookup. Timg factors the per-word write side effects (T0_UPDATE/T0_LOAD, INT_CLR W1C, RTCCALICFG completion, WDT feed/wprotect) into a shared `apply_write_side_effects` helper so byte and word writes produce identical observable state. RtcCntl gets a `read_u32` override only — its OPTIONS0 SW_SYS_RST trigger is intricate enough that the safer choice was to leave `write_u32` on the default byte path (the bench reads RtcCntl but doesn't write it on the hot path).

3. **`Dport` register storage: `HashMap<u32, u32>` → `Box<[u32; 1024]>`** (`crates/core/src/peripherals/esp32/dport.rs`).
   DPORT is exactly 4 KiB / 1024 words. A heap-boxed flat array is the same memory order as a HashMap entry minus the SipHash work. Snapshot wire format stays sparse `Vec<(offset, value)>` so on-disk snapshots interoperate across the layout change. Timg is **not** converted because its `T0_LO`/`T0_HI`/`T1_LO`/`T1_HI` reads rely on HashMap presence-vs-zero distinction to fall back to the live counter — preserving that with a flat array would need a side bitmap, which isn't worth the extra surface for this PR.

Cycle-accurate observable behavior is preserved end-to-end (lockstep / determinism tests pass on default + jit features).

## Bench numbers (ereader, 10M cycles + 200k warmup, mean of 2 runs)

| Stage | Baseline (JIT off) cyc/s | Notes |
|---|---:|---|
| `main` tip `aa016d8` | **284,351** | reference |
| + Fix 1 + Fix 2 (post c628873) | **463,275** | +63% |
| + Fix 3 (this PR tip 50be76c) | **484,693** | **+70% / 1.70× end-to-end** |

The +49% / +27% deltas from the original profiler experiment were quick-and-dirty stubs; the clean implementations land at +63% combined for fix 1+2 and +70% with the array stretch. Below the ×2.44 the profiler projected, but a real, lockstep-clean speedup measured on the same harness.

The ``with browser JIT`` numbers are dominated by JIT compile/refusal overhead and are not the focus of this PR.

## Test plan
- [x] `cargo build -p labwired-core --lib` clean
- [x] `cargo build -p labwired-core --lib --features jit` clean
- [x] `cargo test -p labwired-core --lib` — 529 passed
- [x] `cargo test -p labwired-core --lib --features jit` — 541 passed (lockstep / JIT-feature tests)
- [x] `wasm-pack build --target nodejs --release crates/wasm` clean
- [x] `node scripts/bench_browser_jit.mjs` — numbers above
- [x] Pre-existing failures (`runtime_snapshot::agentdeck_snapshot_file_restores_post_paint_panel`, `labwired-dap` build) reproduce on baseline `aa016d8` and are unrelated to this branch.

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -585,10 +585,25 @@ pub struct Machine<C: Cpu> {
     pub last_breakpoint: Option<u32>,
     pub total_cycles: u64,
     pub config: SimulationConfig,
+    /// Cached bus index of the (single) RTC_CNTL peripheral, resolved at
+    /// construction. `step()` drains the SW_SYS_RST latch every cycle; the
+    /// pre-cache code walked the full ~40-peripheral list and downcast each
+    /// to `RtcCntl` per cycle. With the index cached this collapses to one
+    /// indexed access + one downcast. `None` for configs that don't register
+    /// an RTC_CNTL peripheral (every non-ESP32-classic target).
+    rtc_cntl_index: Option<usize>,
 }
 
 impl<C: Cpu> Machine<C> {
     pub fn new(cpu: C, bus: bus::SystemBus) -> Self {
+        let rtc_cntl_index = bus.peripherals.iter().position(|p| {
+            p.dev
+                .as_any()
+                .and_then(|a| {
+                    a.downcast_ref::<crate::peripherals::esp32::rtc_cntl::RtcCntl>()
+                })
+                .is_some()
+        });
         Self {
             cpu,
             cpu_secondary: None,
@@ -598,6 +613,7 @@ impl<C: Cpu> Machine<C> {
             last_breakpoint: None,
             total_cycles: 0,
             config: SimulationConfig::default(),
+            rtc_cntl_index,
         }
     }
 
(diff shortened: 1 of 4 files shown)


w1ne:
Independently reproduced on real `esp32-epaper-lab` firmware (fixed 50M-step run, release, best of 3):

| Branch | Time | Throughput |
|---|---|---|
| `main` | 41.6 s | 1.20 MIPS |
| this branch | 33.5 s | **1.49 MIPS** |

**~24% faster**, <2% variance. Full suite: 529 passed, 0 failed. Landing.

PR #242, opened 2026-06-11
GPDMA linked-list mode + Cortex-M TBB/TBH base fix (ST DMA_LinkedList twin) + main build fix

w1ne:
Third foreign-firmware twin: ST's **DMA_LinkedList** HAL example (3-node GPDMA queue via `HAL_DMAEx`) reaches its success loop in the simulator and on the bench NUCLEO-H563ZI — same ELF, LEDs observed live over SWD.

## GPDMA linked-list mode (was documented not-modeled)
Bench-probed **first** (capture12: raw 2-block chain driven over SWD — both blocks byte-exact, CLLR→0, BNDT=0, CSR=0x301, CCR keeps TCIE with EN cleared at list end), then modeled:
- node fetch from `{CLBAR[31:16], CLLR.LA}` implemented as memory→own-register DMA copies (one word per CLLR update flag, register order)
- list-start with CBR1=0 (HAL's pattern — first node comes from memory)
- TCF per block (TCEM default), EN-clear only at end of list
- unit test mirrors the silicon capture

## Cortex-M TBB/TBH base fix
The branch-table base for Rn=PC is the **unaligned** in-execution PC (insn+4, ARMv7-M A7.7.185/186); the old `Align(PC,4)` read tables 2 bytes early at 2-mod-4 addresses. Caught by `HAL_DMA_RegisterCallback`: its TBB switch dispatched every callback ID into the same slot, so the error callback overwrote the completion callback and the example "errored" on a clean transfer. The three ARM trace-drift fingerprints legitimately drifted (corrected paths) and are re-baselined with all behavioral assertions green — ledger #6 precedent.

## Main build fix (semantic merge conflict)
`main` currently fails `cargo test -p labwired-core --lib` compile: #237 branched before #239/#240 added `ChipDescriptor.memory_regions` / `SystemBus.extra_mem`, and the squash-merges combined into test literals missing the new fields (each PR was green alone). Fixed here.

## Verification
- DMA_LinkedList twin: success loop on sim AND board (LED1 toggling, error LED dark)
- core lib 1212 green (incl. new linked-list chain test pinned to capture12)
- h563 conformance 6/6 · h563 silicon diffs strict 3/3 on the bench
- tier1 release matrix + ratchet green · trace-drift gate green after re-baseline
- fmt + clippy clean

Ledger #42–43; capture12 archived in the private corpus.


w1ne:
Note: the fdcan commit also *rescues* the FDCAN register model that was stranded on the deleted `h563-foreign-firmware-twin` branch (eb04139f was pushed there after #237 had already squash-merged, so its content never reached main). The model is included here with its IR w1c defect fixed and all three of its unit tests passing. The FDCAN UDS example docs from that commit are still stranded — recover separately if wanted: the old branch head was eb04139f.

w1ne:
Backed the fdcan rescue out of this PR: clippy caught that the rescued register file placed IR at 0x080 and derived RXGFC as 0x0C0 — colliding with TXBC. The map was never CI-run or silicon-verified. FDCAN returns in a dedicated PR, rebuilt against RM0481 and pinned on the bench first, together with bus wiring and a tier1 `can` class. This PR stays scoped to GPDMA linked-list + the main test-suite repair.

PR #244, opened 2026-06-11
test(conformance): standardized per-chip conformance scoreboard + ratchet

w1ne:
**One uniform regression standard for the whole fleet** — instead of bespoke per-chip tests (`esp32c3_reset_conformance`, `h563_mmio_diff`, `stm32f1_exec_oracle`, …). Turns *"are the gates enough?"* into a number per chip, and prevents silent regressions.

## The standard (`chip_conformance.rs`)
The **same battery for every chip** (enforced — a chip can't have a config without being on the board):
- **L0 estate** — descriptor loads; every wired peripheral window is reachable (no bus fault).
- **L1 registers-vs-silicon** — % of a committed real-chip capture (`reg_oracle.json`) the sim reproduces. The deep per-register gate stays in `*_reset_conformance` / `register_coverage`; here we track the headline % so it can't fall.
- **L2 behavior** — a running-firmware gate exists (`firmware_survival` / `*_exec_oracle`), which boots real FW and asserts its register/IO effects.

**Ratchet** (`docs/coverage/chip-conformance.json`): CI fails if any chip's estate breaks, level drops, or silicon match% falls. **Board**: `docs/coverage/chip-conformance.md`. Runs in `cargo test --workspace` — already a gate, no CI yaml change.

### Current board
`esp32c3` is the **reference at L2** (81% silicon reg-match + behavior gate). 9 chips at L1. `esp32 / esp32s3 / g474 / wb55 / wba52` at **L0** — the gaps are now visible red cells, not hidden.

## HW capture infrastructure (run against connected boards on this machine)
- `scripts/hw-oracle/hw_conform.sh <target.json> <out>` — uniform reset-state capture over the debug transport into the standard `reg_oracle.json` schema. **openocd-esp32 transport implemented and proven against the wired C3 (208 words / 6 windows)**; ST-Link SWD / CMSIS-DAP slot in as new branches.
- `scripts/hw-oracle/targets/esp32c3.json` — per-chip target descriptor.
- `README_conformance.md` — the standard + the capture-from-connected-board workflow.

Continuous HIL nightly is a separate, later step (tomorrow); this path runs on demand against locally-connected chips.


w1ne:
Superseded by #245, which merged this exact scoreboard commit as its base along with the STM32L073 silicon wiring on top. Closing as already-in-main.

PR #282, opened 2026-06-13
feat(esp32c3): event-driven virtual AP — deterministic assoc + full DHCP DORA

w1ne:
Replaces timer-blind frame injection with an **event-driven virtual AP** that responds to what the STA actually transmits — making 802.11 association and DHCP **deterministic** instead of flaky, over the real MAC.

## What
- `ap_respond(tx)`: probe-req→probe-resp, auth→auth-resp, assoc-req→assoc-resp, DHCP→DORA, ARP→reply for the **gateway only** (never the STA's own offered IP, which would trip lwIP `dhcp_arp_reply→dhcp_decline`).
- RX reliability: unicast responses delivered at the **front** of the RX backlog (`queue_rx_priority`) so offer/ack land inside the driver's per-state timeout window; beacons only when the RX queue is drained (`pending_rx_len`). Fixes the offer-behind-beacon-backlog stall that left DORA stuck at 'offer'.
- Env-gated traces: `LABWIRED_BRIDGE_TRACE` (frame classifier) + `LABWIRED_DHCP_TRACE` (PC-watch on lwIP `dhcp_*`).

## Verified
Unmodified ESP-IDF C3 binary, real ROM boot: scan→**connected (aid=1, ch6)**→**full DORA (discover→offer→request→ack)**→enters lwIP DHCP CHECKING (ARP self-probe)→advances tries on the fine timer toward `dhcp_bind`. GOT IP bind is fine-timer-paced (follow-up: systimer tick-rate calibration to speed it up).

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1065,6 +1065,171 @@ fn build_dhcp_reply(xid: [u8; 4], ack: bool) -> Vec<u8> {
     f
 }
 
+/// Probe response: like a beacon but unicast to the STA (subtype 5 → FC 0x50),
+/// answering an active scan's probe request.
+fn build_probe_resp(ssid: &str, channel: u8) -> Vec<u8> {
+    let mut f = build_open_beacon(ssid, channel);
+    f[0] = 0x50; // mgmt subtype 5 = probe response
+    f[4..10].copy_from_slice(&BRIDGE_STA); // addr1 = the scanning STA (unicast)
+    f
+}
+
+/// Parse a captured TX data frame as ARP: returns (oper, sender_ip, target_ip)
+/// if it carries an ARP packet (ethertype 0x0806 after LLC/SNAP).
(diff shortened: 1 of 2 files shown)


w1ne:
✅ **GOT IP confirmed end-to-end.** Full 480M-step run with `LABWIRED_DHCP_TRACE=1`:
```
dhcp_check    214.1M  (CHECKING entry, tries→1)
dhcp_fine_tmr 283.4M  → dhcp_check (tries→2)
dhcp_fine_tmr 363.4M  → dhcp_timeout → dhcp_bind   ← BIND
I (15293) probe: GOT IP
```
Unmodified C3 IDF binary bound 192.168.4.2 over the real MAC. Bind is paced by the lwIP 500ms fine timer; it currently fires at ~80M-step intervals due to a systimer period-alarm tick-drop (separate follow-up to speed it ~3x + improve fidelity).

Issue #187, opened 2026-06-07
esp-hal apps crash under fast-boot when the real ROM is provisioned


w1ne:
With the ROM bins vendored (#180), `configure_xtensa_esp32s3` always loads the faithful ROM ("zero thunks") — and esp-hal app startup under FAST-BOOT then dies at `pc=0x40000300 → Exception cause=0 at pc=0x0` (esp-hal calls `rom_config_instruction_cache_mode` @0x40001a1c, which ends jumping to 0 when the ROM was entered mid-stream rather than from reset).

- rom-boot path: unaffected (tier1 esp32s3 row, brom smoke — green).
- e2e fixtures lane: worked around via `LABWIRED_ESP32S3_FASTBOOT=1` (thunk path) — those tests verify firmware logic, not boot fidelity.
- Real fix: make fast-boot under a provisioned ROM viable (pre-initialize the ROM state the cache-config routine expects, or intercept that ROM call in fast-boot mode).

Evidence: nightly run 27106553533 (ESP32-S3 Fixtures E2E), original finding in the tier1 S3 fixture report (2026-06-07).

PR #413, opened 2026-06-28
examples(h563-uds): real-bus demo + decoded-trace autocheck + nightly example smokes

w1ne:
## Why
The `stm32h5-uds-ecu` web playground demo (linked from the *Running a full ECU diagnostic check with no hardware* blog post) showed **no UDS data in the logic analyzer**. Root cause: the demo shipped a stale scripted **loopback** firmware that fails FDCAN init on the current silicon-pinned H5 model — it prints `UDS_FAIL` and halts in a `while(1)`, so **zero** CAN frames ever reach the FDCAN trace the analyzer decodes.

## What
Switch the example to the **real two-node CAN bus** that already exists in this repo: the open ISO-TP responder firmware answering the scripted `uds-tester` device declared in `system.yaml`. No loopback, no fake tester.

- Commit the built responder ELF (`examples/h563-uds-ecu/firmware/h563_uds_ecu.elf`) and point the smokes at it so they run without a cross toolchain (same pattern as the Leo example).
- **`crates/cli/tests/e2e_h563_uds.rs`** — loads the demo, runs the sim, and asserts the FDCAN trace (the exact data `fdcan_trace_snapshot()` feeds the web LogicAnalyzer) carries the `0x22 F190` request **and** the `0x62 F190` positive response. Verified it **fails on the stale ELF** ("never reached ECU_READY") and **passes on the fix**. Runs in the `cargo test --workspace` gate.
- **`scripts/example_smokes.sh`** — runs every example whose firmware ELF is available, asserts each, and reports the *uncovered* ones (no committed ELF) so the coverage gap is visible, not silent. Skips the expected-fail fidelity-bench cases (self-managed by `run-benchmark.sh`).
- **Nightly** (`core-nightly.yml`) — rebuilds the responder from source against pinned UDSLib v2.0.0 and runs the example smokes + e2e, so a firmware / library / model regression fails CI instead of the live blog.

## Verification
- `cargo test -p labwired-cli --test e2e_h563_uds` → pass; against the old ELF → fail.
- `labwired test --script examples/h563-uds-ecu/uds-smoke.yaml` → pass (full scripted diagnostic session: 0x22, multiframe 0xF1A0, NRC, session control, write/read-back, DTC, ECUReset).

## Follow-up
Most other examples are **uncovered** (cargo-built ELFs not present without a build; no e2e). The new `example_smokes.sh` + inventory are the scaffold to close that gap.

---

### Added: ESP32 e-reader de-stub (UC8151D panel + real dual-core)

Same "make the playground examples real, not stubbed" effort.

- **`fix(esp32-ereader)`** — the `labwired-ereader` board declared an `ssd1680_tricolor_290`, but the `GxEPD2_290_C90c` firmware drives **UC8151D** opcodes (PSR `0x00` / PON `0x04` / DTM1 `0x10` / DRF `0x12` / DTM2 `0x13`). SSD1680 reads `0x12` as SWRESET, so 19k SPI transactions reached the panel but `refresh_generation` stayed 0 and it never painted. Switched `system.yaml` + the native e2e to `uc8151d_tricolor_290`; the test now paints (refresh_gen=1, 1429 ink bytes).
- **`feat(esp32-wasm)`** — replaced the single-core handshake-forging stub with a genuine `XtensaLx7::new_app_cpu()` via `with_secondary_cpu`. `loopTask` now runs on a real core 1; the DPORT delivers the cross-core IPI through `Machine::step`. Forging/repin/IPI-bridge kept only as the no-secondary-CPU fallback. Verified in-browser: `get_uc8151d_framebuffer('epaper')` -> 1500 inked bytes on the real second core.

Related: #187 (ESP32 fast-boot fidelity).

### Added: real ESP-IDF heap (dropped the bump-allocator stub)

The firmware's real `multi_heap` (TLSF) allocator now runs on emulated DRAM — the `heap_caps_*` bump-allocator thunks are gone. The years-old "real heap walls" symptom (`heap_caps_malloc` returns a pointer whose first word is the rodata bytes "lock" `0x6b636f6c`, APP_CPU faults on `vector_desc->next` in `esp_intr_alloc`) turned out **not** to be an allocator bug — it was APP_CPU dual-core bring-up. With the genuine second core delivering the DPORT IPI through `Machine::step`, the real heap registers and allocates cleanly. The e2e test paints identically on the real heap (refresh_gen=1, 1429 ink; 83 thunks vs 88), and the playground wasm paints in-browser with no heap thunks.


w1ne:
Follow-up fix: enable the FDCAN1 bus clock in the responder firmware.

After rebasing onto core `main`, the demo regressed to an empty logic-analyzer trace even though the firmware printed `ECU_READY`. Root cause: core `main` added silicon-accurate clock-gating to the H563 `fdcan1` (`clock: { reg: apb1henr, bit: 9 }`, RM0481 §11.8.38). The firmware never enabled `RCC_APB1HENR.FDCAN1EN`, so out of reset the FDCAN register surface read 0 and ignored writes — the `while (CCCR & INIT)` spin saw INIT=0 and printed `ECU_READY`, but the peripheral stayed held in init and rejected every incoming frame.

Fix enables the bus-interface clock at the top of `fdcan_start()` (read-back to order it). The `e2e_h563_uds` test goes red→green on this change, and the browser logic analyzer now decodes the full exchange: ReadDataByIdentifier 0xF190 → "LABWIRED-H563-UDS", multi-frame ISO-TP (FirstFrame 603B → FlowControl → full response), plus session/DTC/routine/IO/comm/reset services (28 UDS events from 36 CAN frames).

PR #461, opened 2026-07-04
feat(test): settle window + min-steps floor for stop_when_assertions_pass (close false-pass hole)

w1ne:
## Problem
`stop_when_assertions_pass` early-stops a `test` run the instant all runtime assertions pass. The implementation broke the loop immediately on first match — with **no settling window and no minimum-steps floor**. That is a false-pass hole: firmware that emits its acceptance token (`verdict=GOOD`) and then aborts/panics would still certify as `assertions_passed`, because the run stopped and declared success at the first token match. For a simulator whose whole value is "firmware verifiably runs," that reduces the verdict to "did the magic string appear."

Verified on the pre-fix binary: a fixture that prints `OK\n` then executes `bkpt` reported `"stop_reason": "assertions_passed"`.

## Fix
`AssertionsPassed` now means "assertions passed **AND** the machine kept executing a settling window afterward without faulting." Because every fault path breaks the loop with its own `StopReason` earlier in the body, simply not breaking immediately — and continuing to run — lets a print-then-crash surface as the fault reason instead.

- New `TestLimits` fields (serde defaults, plumbed script YAML → resolved limits → `execute_test_loop`):
  - `stop_when_assertions_pass_settle_steps: u64` (default `100_000`)
  - `stop_when_assertions_pass_min_steps: u64` (default `0`)
- `execute_test_loop` latches the first all-pass step (gated by the floor), resets on regression, and only sets `AssertionsPassed` once `step - first >= settle_steps`.

## Test (TDD)
`test_stop_when_assertions_pass_does_not_mask_crash` runs the new fixture `tests/fixtures/uart-then-bkpt-thumbv7m.elf` (writes the token to UART1 then `bkpt #0` → `Halt`) with `settle_steps: 50` and asserts the token IS present yet `stop_reason` is a fault (`halt`), not `assertions_passed`, and `status != pass`. Pre-fix: `assertions_passed` (the bug). Post-fix: `halt`. `cargo test -p labwired-cli --test stop_conditions` 8/8; `cargo test -p labwired-config` 22/22; fmt clean.

## Residual limitations (honest — read before relying on this as a prod gate)
1. **Only faults that break the loop are caught.** The cortex_m core *swallows* bad loads/stores and undefined instructions (fidelity-logged, not propagated as `MemoryViolation`/`DecodeError`). The only firmware-reachable fault that breaks the run is a trap/abort (`bkpt`, which a Rust `panic!`/abort lowers to). So the window closes the hole for **aborting/trapping** firmware and for **print-then-hang/timeout**, but a print-then-*untrapped-wild-write* is still out of scope. Follow-up: make the settle window also fail on fidelity-logged memory faults, or propagate them.
2. **A clean `SYSRESETREQ` reboot after the token is NOT caught** — UART accumulates across the reset so assertions keep matching post-reset and the reset itself doesn't fault.

## Context
This is the core half of proto.cat's STM32 stop-on-evidence fix (protocat-private#97). The builder + proto.cat still need to pass the two new limit fields through; the deployed builder binary is also stale. Not for merge until CI is green and the residual-limitation call is made.


w1ne:
### Real-firmware proof (STM32L476, not just the synthetic fixture)

Ran the committed `stm32l476-zephyr-hello.elf` (Zephyr `hello_world` on nucleo-l476rg) through the new binary with the de-risked stop:

```
limits:
  max_steps: 45000000
  stop_when_assertions_pass: true
  stop_when_assertions_pass_settle_steps: 100000
  stop_when_assertions_pass_min_steps: 1000
assertions:
  - uart_contains: "Hello World! nucleo_l476rg"
```

Result — clean early stop in 1.3s:
```
status:        pass
stop_reason:   assertions_passed
steps_executed: 111111        # banner printed ~11k + 100k settle window
uart.log:      *** Booting Zephyr OS build c66235fb7346 ***
               Hello World! nucleo_l476rg/stm32l476xx
```

This is the exact proto.cat STM32 failure mode: the firmware **runs and prints correctly**, but the hosted builder posts a fixed `max_steps: 45_000_000` against a 60s wall-clock with no stop-on-evidence, so a run that's *done* at ~111k steps grinds until it times out. With the settle window it stops at 111,111 steps, `assertions_passed`, having survived 100k steps past the banner without faulting — so the token means "reached AND survived," not just "appeared."

PR #504, opened 2026-07-10
Standardize the external stimulus API: SimInput on every input device + channels in the device schema

w1ne:
## What

Makes the generic input/stimulus API the ONE standard surface for driving external signals into simulated devices, and publishes each device's drivable channels as part of the device schema.

Before: `Machine::set_input` / `list_inputs` / test-script `stimuli:` / MCP `run_lab` could reach exactly one device (FXOS8700, I2C-only walk). Everything else needed a bespoke WASM setter.

After: every input-capable model is discoverable and drivable by channel name through the same API, from YAML, MCP, WASM, or Rust:

| device | transport | channels |
|---|---|---|
| FXOS8700 | I2C | x, y, z (g) |
| ADXL345 | I2C | x, y, z (g) |
| MPU6050 | I2C | ax, ay, az (g), gx, gy, gz (°/s) |
| VL53L1X | I2C | distance (mm) |
| MAX31855 | SPI | temperature, internal (°C) |
| 74HC165 | SPI | ch0–ch7 (level) |
| NEO-6M GPS | UART | lat, lon (°), fix (level) |
| HC-SR04 | GPIO/bus | distance (cm) |

## How

- `SimInput` impls on all 7 new models; `as_sim_input_mut` hook added to `SpiDevice` + `UartStreamDevice` (same contract as `I2cDevice`), forwarded through `TracingSpiDevice`.
- One `for_each_sim_input` walk behind both `list_inputs` and `set_input`, covering I2C devices, SPI devices, UART streams, and bus-direct HC-SR04 (reported under its sensor id).
- `StimulusTarget.component` is now real, not advisory: narrows resolution to the named owner, accepting the external-device id from system.yaml (alias table recorded at attach) or the owning peripheral bus name. `Machine::set_input_on` exposes it; the CLI stimulus runner passes it through. Collisions without `component` stay a typed `Ambiguous` error.
- **Schema**: `KitMetadata.inputs` — the SAME static channel table backs the `SimInput` impl and the manifest entry, so `peripherals-manifest.json` now advertises each device's drivable channels (key/label/unit/min/max) and cannot drift from the runtime. Manifest fixture regenerated (additive field, schema_version unchanged).

## Known gaps (unchanged behavior)

- NTC thermistor: seeded into the ADC once at attach, not bus-resident — stays on the bespoke WASM path until ADC channel sources are modeled as devices.
- Raw ADC injection, UART byte feed, and GPIO button presses remain WASM-only surfaces (not channel-shaped).
- FXOS8700 + HC-SR04 are not kit-registered, so they appear in runtime `list_inputs` but not the static manifest.

## Tests

- New matrix board test (STM32F103 with devices on I2C/SPI/UART/GPIO): per-transport driving with model-level readback, ambiguous-key errors, component disambiguation by peripheral name AND by external-device id.
- `e2e_kw41z_cow_stimulus` green — its `component: fxos8700` now actually resolves (previously ignored).
- `peripheral_kit_gate`, config, clippy, fmt, wasm32 build, validation drift gate all green. The 3 failing `cpu::riscv` WFI lib tests fail identically on clean origin/main (pre-existing).

The superproject needs `packages/ui/src/peripherals/manifest.json` regenerated when this lands (gate test compares both).


w1ne:
Pushed a second commit addressing self-review findings: (1) input conversions now follow the live full-scale config registers (MPU6050 ACCEL/GYRO_CONFIG, ADXL345 data_format, FXOS8700 xyz_data_cfg) with silicon-accurate saturation, (2) per-device identity stamped at attach — discovery and `component` now both speak the system.yaml id, and two same-type devices on one bus are individually addressable (alias table removed), (3) atomic `set_inputs` batch (validate-all-then-apply) so multi-channel poses are never observed torn.

PR #533, opened 2026-07-13
jit(riscv): wasm emit skeleton + integer-ALU codegen (chunk C)

w1ne:
## RISC-V JIT chunk C — wasm emit skeleton + integer-ALU codegen

First frontend block that **executes instead of bailing**. Builds the shared emit infrastructure (wasm module builder, guest register-file marshalling, wire-code→SideExit return protocol) that chunks D (branches) and E (load/store) layer onto, and emits real wasm for the full RV32IMC integer-ALU surface.

### What's emitted
- **Reg-imm / reg-reg arith + logic**: `lui, auipc, addi, slti[u], xori, ori, andi, add, sub, sll, slt[u], xor, srl, sra, or, and` and the immediate shifts.
- **RV32M**: `mul, mulh[su|u]` (i64-extend → `i64.mul` → `shr_u 32` → wrap), and trap-free `div[u]/rem[u]` with explicit guards for div-by-zero and the `INT_MIN/-1` signed-overflow case (wasm `div_s`/`rem_s` trap where RISC-V defines a result — every case guarded to match the interpreter).
- **Compressed**: `c.addi, c.li, c.mv, c.addi16sp, c.addi4spn, c.slli`.

### Model
Register-file-in-locals: prologue loads only the registers the block *reads* from an imported wasm memory into locals, the body operates purely on locals (`x0` reads as const 0, writes drop), the epilogue stores only *written* registers back. Fixed 128-byte host sync per block regardless of length. The executor is RISC-V-local (the generic `JitRuntime::run` deliberately elides the register-file handle) and drives `Machine<RiscV>` one dispatch unit at a time via `step_unit`.

### Correctness gate
`riscv_jit_alu_lockstep.rs`:
- **byte-identity** — a hot ALU loop (incl. div/rem guarded paths driven to their edge cases every iteration) run through compiled blocks is StateVec-identical to the interpreter, non-vacuously (`compiled > 0`, `block_runs > 0`, instruction floor).
- **per-op differential** — each ALU op as a one-instruction compiled block vs. the interpreter over random operands.
- **micro-bench** — compiled ALU loop vs. interpreter, reports the measured ratio.

Not yet wired into production dispatch — test-only until chunk H's JitHost/SafetyGate adapter. Full regression green (`jit,event-scheduler` + `jit-framework`), fmt + clippy clean.

Base: `jit-framework-scaffold` (do not merge to main until D/E land and the ≥3× invaders MIPS bar is met).

### DIFF
--- a/crates/core/src/cpu/jit_framework/riscv/emit.rs
+++ b/crates/core/src/cpu/jit_framework/riscv/emit.rs
@@ -1,0 +1,540 @@
+// LabWired - Firmware Simulation Platform
+// Copyright (C) 2026 Andrii Shylenko
+// SPDX-License-Identifier: MIT
+
+//! RV32IMC integer-ALU wasm codegen (JIT framework chunk C).
+//!
+//! This is the **first block that executes instead of bailing**. It turns a
+//! straight-line run of integer arithmetic / logical / shift / mul-div
+//! instructions into a real wasm module (via [`super::wasm_encode`]) that a
+//! runtime ([`super::exec`]) runs against the guest register file.
+//!
+//! ## The register-file-in-locals model
+//!
+//! The compiled module imports one memory — the **guest register file**,
+//! word `i` = `xi` at byte offset `i*4` (see
+//! [`wasm_encode::REGS_IMPORT_MODULE`]). Each block:
+//!
+//!   1. **Prologue** — loads every register it *reads* from that memory into
+//!      a wasm local (`x0..x31` map to locals `0..31`).
+//!   2. **Body** — one emit per guest instruction, operating purely on
+//!      locals. `x0` reads as the constant `0` and writes to it are dropped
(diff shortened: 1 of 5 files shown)


w1ne:
**Micro-bench (native wasmtime, pure-ALU hot loop):**
```
interp: 1.2 MIPS
jit   : 20.2 MIPS
speedup: 16.86x   (compiled_blocks=64 block_runs=7024 block_instrs=449536 interpreted=7283)
```
Byte-identity: `alu_hot_loop` retired 5584/6000 instrs through compiled blocks, StateVec-identical to the interpreter incl. the guarded div/rem edge cases. Per-op differential green over random operands.

This is the best-case ceiling (no branches/loads/MMIO — those bail to the interpreter until chunks D/E). It confirms the emit path itself has large headroom; the real ≥3× **invaders** MIPS bar is a whole-program number that D+E gate.

PR #558, opened 2026-07-16
perf(sim): coalesce millis/SYSTIMER busy-polls under idle FF

w1ne:
## Why
Web C3 workshop stays **super slow** after tick-512: the Arduino sketch busy-polls `millis()` (SYSTIMER freerunning snapshot), so **idle FF never fires** (no WFI). Host ~45–50 MIPS busy ⇒ ~0.3× wall; browser wasm is slower still.

## What
Under existing `idle_fast_forward` (already default ON for C3 in playground):

- Track per-batch MMIO via **CPU-agnostic** `MmioAccessClass` (`FreerunningTimerPoll` / `SideEffecting` / `SideEffectFree`) on the `Peripheral` trait — **no chip names in SystemBus**
- Ignore XIP/flash windows (side-effect-free rodata)
- If batch is pure timer-poll (≥2 snapshot accesses, zero other MMIO), skip to next **scheduler event** (FreeRTOS tick, etc.)

**Not fake millis** — `cycle_clock` / SYSTIMER advance with skipped device cycles; only empty spin is not interpreted.

## Architecture follow-ups in this PR
- `bus/mmio_activity.rs` — class tallies only
- `bus/can_devices.rs` + `bus/faults.rs` peeled from `bus/mod.rs`
- SYSTIMER offsets: `@generated` `systimer_regs.rs` from checked-in `esp32s3.svd` (`scripts/gen_esp_systimer_regs.py`); offline test pins FREERUNNING_POLL + key regs

## Tests
- `timer_poll_coalesce_uses_peripheral_access_class_not_chip_names`
- `generated_offsets_match_svd`
- bus suite (CAN/UDS/faults) green after peel

## Ship
After merge (no admin merge from agent): monorepo pin + wasm rebuild (same as #557). Docs only in monorepo parent if needed.


w1ne:
**Architecture fix (CPU-agnostic core):**

Shared `SystemBus` no longer hardcodes `"systimer"` names or ESP register offsets.

- New `MmioAccessClass` on `Peripheral` (default **SideEffecting**)
- ESP knowledge stays in `Systimer::mmio_access_class` / `FlashXipPeripheral::mmio_access_class`
- Bus only tallies classes → one chip optim cannot influence another by name

Machine idle-FF still only skips when a batch is pure freerunning-timer polls.

w1ne:
**Follow-up refactor (your feedback):**

1. **No ESP magics on the bus** — `SystemBus` only counts `MmioAccessClass`.
2. **SYSTIMER offsets derived once** — `systimer::regs::{UNIT0_OP, …}` + `FREERUNNING_POLL` array used by both `read_word`/`write_word` and `mmio_access_class`.
3. **Out of mod.rs** — bookkeeping lives in `bus/mmio_activity.rs`.

Next optional step: keep peeling more `bus/mod.rs` into focused modules (routing/tick already exist).

PR #643, opened 2026-07-24
test: un-rot the two long-red strict_onboarding + xtensa_exec integration tests

w1ne:
Fixes two pre-existing, long-red integration tests that were excluded from PR CI and rotted (they fail locally and in the full-CI lane). Two independent root causes, one commit each. No production/engine code changes — test-only plus new example metadata.

## 1. strict_onboarding — `unexpected example gaps: ["nrf54l15"]`

**Root cause:** `find_example_for_chip` (crates/core/tests/strict_onboarding.rs) associates a chip with an example dir by scanning `examples/*/system.yaml` for a `chips/<chip>.yaml` reference. The two real nrf54l15 examples (`nrf54l15-dk`, `nrf54l15-smart-ring`) shipped their board manifests under `configs/systems/` (`nrf54l15dk.yaml`, `smart-ring.yaml`) and had **no local `system.yaml`**, so the matcher found nothing and reported a gap.

**Fix (association made real, no skip/allow-list entry):**
- Added a local `system.yaml` to each example dir, mirroring `configs/systems/` — the same convention every other onboarded board (nucleo-f401re, demo-blinky, …) already follows.
- Added an `io-smoke.yaml` to each that runs the committed bare-metal fixture and asserts the real, silicon-sourced tokens: the DK smoke asserts the UARTE20 boot banner (`nRF54L15 boot OK`, core/RRAM/RAM geometry, register bases); the smart-ring smoke asserts every TWIM21 sensor (BMI270/MAX30102/TMP117/DRV2605) answers its WHO_AM_I over the modelled I²C bus.
- Taught `ensure_smoke_firmware_exists` to accept committed-fixture firmware: the nrf54l15 fixtures are pre-built ELFs under `tests/fixtures/` (external `arm-none-eabi-gcc`), not cargo `target/` artifacts, so it now requires their presence instead of returning a spurious build failure.

Verified assertions are non-vacuous (a deliberately wrong `uart_contains` fails the runner with exit 2). Gate now prints `[PASS] nrf54l15 is strictly onboarded`.

## 2. xtensa_exec — 2 MOVSP window-exception failures

**Root cause (regression, with bisect evidence):** The tests (`test_exec_movsp_overflow_raises_exception`, `test_general_exception_movsp_alloca_redirects_pc`) were added in #80 when MOVSP unconditionally raised AllocaCause (EXCCAUSE=5) on a live adjacent frame. `0643a5b0` gated the faithful vector path behind `faithful_windows` but deliberately kept a hard `raise_general_exception(5)` in fast-boot mode *"to restore the non-faithful MOVSP exec tests."* Then **#618 (`41da5d27`)** removed that fast-boot hard-fault and replaced it with a plain register move — because vectoring mid-boot into an unprimed handler hard-faults ESP-IDF `heap_caps_init`'s VLA/alloca path. That was a correct real-firmware boot fix, but it silently invalidated the two tests.

**Fix:** The ISA-faithful AllocaCause semantics still live in `faithful_windows` mode (`vector_exception(5)`: EPC1←PC, EXCCAUSE←5, PS.EXCM←1, PC←VECBASE+0x300, then continue into the handler). Restoring the fast-boot hard-fault would re-break real boot (violates the never-break-existing-chips gate), so the two tests are retargeted at faithful mode — where the Xtensa RM §8 MOVSP spill/AllocaCause behavior is genuine — and now assert the full vector state. Assertions are strengthened, not weakened; no production code changed.

## Verification
- `cargo test -p labwired-core --test strict_onboarding` → 1 passed (`[PASS] nrf54l15`).
- `cargo test -p labwired-core --test xtensa_exec` → 194 passed, 0 failed, 3 ignored (was 192/2).
- `cargo fmt` clean; `cargo clippy -p labwired-core --tests` clean.

The remaining full-suite reds (`e2e_stm32f407_i2c`, `firmware_survival::test_nucleo_{f407,l476rg}_i2c_survival`) are the STM32 I²C-fleet domain owned by a separate stream; confirmed failing identically on pristine current `main` and untouched here.


w1ne:
Heads-up: this PR's two fixes are complete and green in isolation (`strict_onboarding` 1 passed → `[PASS] nrf54l15`; `xtensa_exec` 194 passed / 0 failed; fmt + clippy clean on the two touched test files).

It is currently blocked only by pre-existing **main-integrity** failures in the pr-gate lane that are unrelated to this changeset and live in the fenced-off I2C/declarative domain — confirmed failing identically on pristine `origin/main` (`46bf272b`):

1. **Check Formatting** — `crates/core/src/peripherals/components/declarative_regs.rs:101,109` is unformatted (SPI-L3 declarative migration, #638/#641).
2. **Validation drift** — `esp32s3` has drifted (model 2026-07-24 > capture 2026-07-15, no covering `drift_ack`) because #644 (central I2C `advance_time_us`) touched `esp32s3/systimer.rs` without bumping esp32s3's `drift_ack` in `validation/manifest.yaml`.

Auto-merge (MERGE) is armed; it will fire once main is green (a `cargo fmt` on declarative_regs.rs + an esp32s3 `drift_ack`/re-capture from the owning streams). I've deliberately not touched either, per the I2C/timing stream boundary.

PR #676, opened 2026-07-26
perf(sim): 12.6x on the shipped ESP32-C3 labs — scheduler hashing + I2C wake cadence

w1ne:
Two hot-path fixes to the simulator, each byte-identical on every gated
workload, plus the CI lane that turned out not to exist.

## Result

End-to-end on the firmware the playground actually boots (median of 8
interleaved rounds, both binaries built identically and `cmp`-verified
distinct, run alternately):

| shipped lab | before | after | speedup | RTF |
|---|---|---|---|---|
| `esp32c3-oled-lab` (128x64) | 2.155 s | 0.171 s | **12.6x** | 0.087 -> **1.098** |
| `esp32c3-oled-128x32-lab` | 1.170 s | 0.106 s | **11.0x** | 0.160 -> **1.769** |

RTF crossing 1.0 means these labs now simulate faster than the silicon
they model. They were ~12x below real time.

Deterministic, machine-independent evidence (the shipped-lab gate asserts
batch width rather than wall time for exactly this reason):

```
esp32c3-oled-lab        batches 175104 -> 27156   mean_batch 19.418 -> 125.212
esp32c3-oled-128x32-lab batches 115496 -> 26527   mean_batch 26.323 -> 114.609
interpreted / lit_px    3400255 / 782  and  3040236 / 442   -- UNCHANGED
```

## 1. `perf(sched)` — the event scheduler was 27% SipHash

`sample`(1) put `sip::Hasher::write` as the top non-idle leaf frame in the
whole simulator, above the RISC-V interpreter. `EventScheduler` hashed a
16-byte key twice per event to maintain a set whose high-water mark is
**3 entries**. `queued` -> linearly-scanned `Vec`, `live_per_peripheral`
-> dense `Vec` indexed by slot. New `SchedulerStats::max_queued_events`
makes the sizing assumption observable instead of assumed.

## 2. `perf(esp32c3-i2c)` — wake per wire segment, not per module tick

A module tick with `ticks_left > 1` does exactly one thing: decrement that
counter. `chase()` only fires at zero, every level change goes through
`enter()` inside `transition()`, and `advance_engine` is delta-based, so N
ticks in one call lands on exactly the state N single-tick calls would.
`TxStall` arms `ticks = 1`, so clock-stretch retry cadence is unchanged by
construction.

**Widening alone is unsafe, and not theoretically.** `CTR.FSM_RST` parks
the engine mid-transaction while a wake stays queued (no cancel API, by
design); the following `TRANS_START` finds `scheduled == true`, arms
nothing, and the stale wake clocks the fresh transaction — measured 108+
cycles late. Arduino's `i2c_ll_master_clr_bus()` writes FSM_RST, so real
firmware takes this path. It is invisible at a 4-cycle cadence, which is
why widening *creates* it. Fixed with `arm_seq` + `re_anchor()`, layer 3
of the scheduler's own documented cancellation contract.

## Gates — both verified to FAIL without their fix

- `rearm_after_fsm_rst_matches_the_per_cycle_walk` — samples SCL/SDA every
  cycle against the legacy per-cycle walk, the reference that has no wakes
  and so cannot be wrong about them. Diverges at cycle 1536 without the
  token fix.
- `every_attached_i2c_device_matches_the_per_cycle_walk` — the whole
  `build_i2c_device` roster (14 devices incl. mlx90640, which had no
  differential at all) through write-then-repeated-START-read with a
  mid-transaction bus reset, comparing waveform + SR + INT_RAW + RX FIFO.
  Fails if the factory gains a device the roster omits.

Both nearly shipped vacuous and are documented as such in-file: the first
passed against broken code until the timing was made realistic (default
reset timing gives 9-tick segments, too short for a stale wake to show);
the second compared `rx=[] == rx=[]` across all 14 rows until the budget
reached the READ phase.

## 3. `fix(ci)` — core had no release lane

`cargo test --release -p labwired-core` **did not compile** on main.
Every cargo step in `core-ci.yml` builds debug, so
`#[cfg(not(debug_assertions))]` in core was not untested — the compiler
never saw it, and `clippy --all-targets` missed it for the same reason.
`tests/event_scheduler.rs` called `drain_due(&[0u32])` against a
0-argument API; its sibling `unbounded_distinct_rearm_is_counted_in_release`
had never executed once. Adds the lane plus
`release_only_cfg_ratchet.rs`, which fails when release-only code appears
in a file that lane does not name.

Also in that commit: `chip_conformance` regenerated its own committed
expectation via `fs::write(..).ok()` and passed, so the board drifted
(nrf52832 reg-match 10 vs a real 16, stm32f407 29 vs 31) — now checks by
default; and `e2e_nokia5110_invaders` asserted a hard-coded
`max_safe_tick_interval == 64` long after the constant moved to 512.

## Verification

**112 suites / 111 binaries, 3052 passed, 16 failed** with
`--include-ignored --no-fail-fast`, re-run after merging `origin/main`
(#673 touches C3 boot) with identical results.

All 16 failures are pre-existing and `#[ignore]`d: 15 die on missing
build artifacts (arduino-matrix ELFs, `/tmp/*.lwrs`, an ereader ELF whose
own message says "NO CI lane builds" it, the Xtensa target).
`xtensa_exec` is a genuine assertion failure — **reproduced on pristine
`origin/main`**, same three tests, same values.

Note: `cargo test -- --include-ignored` never terminated on this crate
before this PR — two `sample`(1) targets looped until killed, so runs
reached 34 of 111 binaries and silently never built the rest. Four of the
failures above were sitting past that hang point, unseen. They now take a
wall-clock budget (`LABWIRED_C3_SAMPLE_SECS`, 10 min default).


w1ne:
### CI provenance before merge

`core-integrity` — which contains the new release-gated lane this PR adds — **passes** on this branch. That lane skips on pull requests, so it was dispatched explicitly rather than assumed:
https://github.com/w1ne/labwired-core/actions/runs/30210736518

`core-full` fails on this branch at `Run Workspace Tests`:

```
labwired-hw-oracle::stm32f1_exec_oracle::exti0_interrupt_delivery_sim
thumb oracle: mem[0x20000300]: expected 0xABCD1234, got 0x00000000
test result: FAILED. 13 passed; 1 failed
```

**This is pre-existing, not introduced here.** The same workflow dispatched on `main` fails at the same step, on the same test, with the same values:
https://github.com/w1ne/labwired-core/actions/runs/30211209195

```
thread 'exti0_interrupt_delivery_sim' panicked at crates/hw-oracle/src/arm_thumb.rs:149:9:
assertion `left == right` failed: thumb oracle: mem[0x20000300]: expected 0xABCD1234, got 0x00000000
test result: FAILED. 13 passed; 1 failed
```

Worth checking separately: it passes locally (isolated and as the full 14-test suite) and fails only on the CI runner, so it looks environment- or ordering-dependent rather than a model defect. `core-full` is dispatch/schedule-gated, so nothing routinely runs it.

Also filed against this branch, all pre-existing and all `#[ignore]`d — they surface only under `--include-ignored`, which nothing in CI passes: `diag_c3_yield`, `diag_esp32{,c3,s3}_*` (6 suites, missing `validation/arduino-matrix/out/*.elf`), `runtime_snapshot` (missing `/tmp/*.lwrs`), `jit_lockstep` (ereader ELF its own message says no CI lane builds), `esp32s3_oled_profile` (missing chip descriptor), and `xtensa_exec` (a real assertion failure, reproduced on pristine `main`).

PR #683, opened 2026-07-27
feat(rp2040): model the PWM block + document onboarding candidates

w1ne:
RP2040 was the weakest-covered popular part in the fleet — **6/12** tier-1 and
one of only two ⚪ `structural` boards — because three of the most-used
peripherals on a Pico (PWM, ADC, RTC) had no model at all and the watchdog was a
bare `stub`. This is the first instalment.

### PWM (datasheet §4.5) — `pwm=na → pass`

Eight slices, modelled behaviourally rather than as a register bank:

- counter wraps at `TOP` and latches `INTR`; **PWM_IRQ_WRAP (NVIC 4)** is held
  level-sensitively while a masked wrap is pending, matching the timer model
- the **fractional divider is a real phase accumulator** — `DIV=2.5` advances on
  alternate pairs of ticks rather than being rounded to 2 or 3
- **phase-correct** ramps up and back down with period `2*TOP`, holding neither
  endpoint for an extra step
- `EN` (0xA0) genuinely aliases the per-slice `CSR.EN` bits both directions,
  which is what makes an atomic multi-slice start work

`DIVMODE`'s gated/edge modes deliberately do **not** free-run — there is no
B-pin routing into this model, so a gated slice has no clock and standing still
is the honest result. The module header states plainly what is and isn't
modelled.

11 unit tests: wrap, IRQ mask/ack, EN alias, integer + fractional dividers,
phase-correct direction, gated inertness, self-clearing `PH_ADV`/`PH_RET`,
datasheet reset values.

### A silent build bug this uncovered

`examples/tier1-fixture/rp2040/build.rs` never emitted `-Tlink.x`, unlike its
nRF siblings. Rebuilding that fixture with the sanctioned
`scripts/tier1/build_nordic_rp2040.sh` recipe produces an ELF with **entry 0x0
that does not boot** — and the build *succeeds*, so it fails silently.

Proven by A/B: committed ELF enters at `0x100000c1` and reports 5 PASS lines; a
plain rebuild enters at `0x0` and emits nothing, turning **every** rp2040 tier-1
cell `blocked`. Anyone regenerating that fixture would have hit this. Fixed at
the source so a plain `cargo build` is correct rather than requiring callers to
remember RUSTFLAGS.

### Also: `docs/onboarding-candidates.md`

Catalogues what to onboard next, tiered by **real cost given models already in
the tree** — Tier A (STM32F411/Black Pill, G431, L432, H723, nRF52833) needs only
a config + fixture; Tier B (RP2350, STM32U5, G0, ESP32-C6) needs new registers
but no new core; Tier C (SAMD21, RA4M1, CH32V003, i.MX RT1062) each open a vendor
the engine has never modelled — Microchip and Renesas are both absent today.
Records what's out of scope and why, and states that onboarding lands a part at
`structural`/`sim-validated`: only 9 of 15 boards carry a live silicon capture.

### Verification

Full matrix diffed against the pre-change baseline — **exactly one cell moved**,
no other chip touched. core lib 2341, firmware_survival 51/51,
board_coverage_ratchet, strict_onboarding all green; rustfmt and clippy clean.

Independent of #677/#678/#680/#681/#682. ADC, RTC and a behavioural watchdog are
the natural follow-ups (rp2040 → ~11/12).


w1ne:
## ADC + RTC + watchdog, and the `irq` class — rp2040 tier-1 7/12 → 11/12

Follow-up to the PWM commit on this branch. The remaining two unmodelled blocks
plus the watchdog, which was a `stub` (pure storage), so firmware could enable
it and poll `CTRL.TIME` forever with nothing happening.

Every address, offset, field position, reset value and IRQ number came out of
`tests/fixtures/real_world/rp2040.svd`, not from a hand transcription. That
caught two wrong numbers in the spec I was working from: **ADC `CS.RROBIN` is
[20:16]**, not [24:16] (five inputs, five bits), and **RTC `IRQ_SETUP_0` packs
`DAY_ENA`/`MONTH_ENA`/`YEAR_ENA`/`MATCH_ENA`/`MATCH_ACTIVE` at 24/25/26/28/29**,
not 26/27/28/29/30.

**ADC** (`0x4004C000`, `ADC_IRQ_FIFO` = 22) — 12-bit SAR, 8-entry FIFO with
LEVEL/EMPTY/FULL/OVER/UNDER and `FCS.SHIFT`, round-robin sequencer, DIV-paced
free-running mode with a real fractional accumulator. `INTR.FIFO` is a *level*
(`LEVEL >= THRESH`), not a latch. The temperature sensor converts the datasheet
relation `T = 27 - (V - 0.706)/0.001721` through named constants, so 27 °C
*produces* code 876 rather than 876 being written down; `TS_EN` gates it and an
unpowered sensor converts 0. Inputs 0..3 are drivable through
`SystemBus::seed_adc_channel` (same seam as the STM32/S3 ADCs), so a pot or
thermistor in `system.yaml` moves a real level — but there is no invented
default, because a fabricated mid-scale reading looks exactly like a working
sensor that is not connected.

**RTC** (`0x4005C000`, `RTC_IRQ` = 25) — a real calendar with carry, per-field
alarm, and **the `RTC_1`-before-`RTC_0` read latch**: read `RTC_0` first and you
get a stale time, here as on silicon. Leap years use the hardware's own
`year % 4` rule (which is why `FORCE_NOTLEAPYEAR` exists), so 1900 is a leap year
to this RTC unless firmware says otherwise.

**Watchdog** (`0x40058000`) — the RP2040-E1 **double decrement**: the counter
falls by 2 per tick, so `LOAD` is in half-microseconds and pico-sdk's
`watchdog_enable()` multiplies ms by 2000. The tick generator really divides, and
`TICK.RUNNING` needs `ENABLE` *and* non-zero `CYCLES` — which is exactly why the
reset state reads `0x200` and not `0x600`. Expiry latches `REASON.TIMER` and
stops the dog rather than resetting the CPU (the Nordic WDT convention).

**irq** — the yaml now declares the NVIC (installed unconditionally for Cortex-M
anyway; the entry just tells the matrix it exists). The fixture check is not a
software pend: it arms PWM slice 1, unmasks `PWM_IRQ_WRAP`, and requires the
vector to actually run.

Each module header states what is *not* modelled, including the one that needed
real thought: the watchdog tick generator and the TIMER model are not on a common
time base (on silicon `TICK.CYCLES` sets both), so with the usual `CYCLES = 12`
they are 12× apart in sim time. Recorded in `watchdog.rs` rather than hidden by
ignoring `CYCLES`.

### Verification

Live fixture: `clock timer gpio spi i2c pwm adc rtc wdt irq` all PASS + `uart`
implicit. Only `dma` remains unrecorded.

Full 16-chip matrix run twice with the same binary shape, pristine HEAD vs this
change — **exactly four cells moved, all rp2040, all `na` → `pass`** (`adc`,
`irq`, `rtc`, `wdt`). No other chip touched. `pwm`'s pass, earned in `deffb92f`
but never snapshotted, is recorded now too.

| suite | result |
|---|---|
| `cargo test -p labwired-core --lib` | 2383 passed, 0 failed (was 2341; +42 new) |
| `cargo test -p labwired-cli --lib` | 39 passed, 0 failed |
| `firmware_survival` | 51 passed, 0 failed, 5 ignored |
| `board_coverage_ratchet` | 2 passed |
| `strict_onboarding` | 1 passed |
| `register_coverage_ratchet` | 1 passed |
| `chip_conformance_ratchet` | 1 passed (board regenerated: rp2040 15 → 19 peripherals, no other row) |
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --all-targets -- -D warnings` | clean |

### Two things found, both out of scope

1. **`stm32g474re` is 7 cells short of its snapshot on this branch and on
   `origin/main`** — `clock-enr`, `tim-arr32`, `pwm-ug-latch`, `i2c-txe-reset`,
   `spi-txe-reset`, `adc-adrdy`, `rtc-dr-reset`. Pre-existing, identical before
   and after this change, and already fixed on `fix/tier1-g474-rcc-offsets`. For
   that reason the snapshot's rp2040 row was updated **in place** instead of
   regenerating the whole file, which would have committed that in-flight
   regression as the new baseline.
2. **`docs/coverage/register-modeling.json` is badly stale** — measured today,
   rp2040 is 722 modelled vs a 67 baseline, esp32c3 2505 vs 333, nrf52832 397 vs
   2. The ratchet only checks `>=` so it passes, but it is currently measuring
   almost nothing. Left alone deliberately.

PR #749, opened 2026-08-02
ci: make the runner a variable instead of 31 hardcoded literals

w1ne:
Every job in this repo pinned `runs-on: ubuntu-latest` directly. When GitHub-hosted runners went unavailable on 2026-08-02 (account billing block), there was no way to move this repo's CI without editing **31 job definitions**.

**Nothing changes today.** The default is still `ubuntu-latest`, so routing is byte-identical and hosted runners keep being used. This only adds the switch: setting the `CI_RUNNER` repo variable now redirects the whole repo at once — the same pattern the superproject already uses.

## Deliberately not converted

- `core-release.yml` → `runs-on: ${{ matrix.runner }}` — the cross-platform release matrix. Forcing it onto one Linux pool would break the macOS/Windows builds.
- `core-hil.yml` → `["self-hosted","hil","${{ matrix.runner_label }}"]` — HIL rigs are physical hardware and already correct.

## Verification

- `scripts/ci/verify-release-runner-contract.sh` **executed against this tree and passes** — it greps `core-release.yml` for literals, so it was the one gate that could plausibly break. 13 tests, OK.
- `actionlint 1.7.7` clean on every file I changed. It does warn about the custom `hil` label in `core-hil.yml:59`, but that is **pre-existing** — identical on `origin/main`, on a line this PR does not touch, and this repo has no actionlint gate.

## ⚠️ For whoever flips the switch

This repo currently has **one** self-hosted runner registered (`pluto-labwired-core-ci`). Setting `CI_RUNNER` serializes all 31 jobs onto it. That's an outage lever, not a steady state.

The billing block has since cleared — hosted runs succeeded at 19:06 and 20:07 — so there is no need to flip it now. This is so that next time it takes one variable rather than a 31-file PR.

### DIFF
--- a/.github/workflows/core-arduino-matrix-smoke.yml
+++ b/.github/workflows/core-arduino-matrix-smoke.yml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   arduino-matrix:
     name: arduino-${{ matrix.board }}
-    runs-on: ubuntu-latest
+    runs-on: ${{ fromJSON(vars.CI_RUNNER || '["ubuntu-latest"]') }}
     timeout-minutes: 45
     strategy:
       fail-fast: false
@@ -117,7 +117,7 @@ jobs:
 
   arduino-matrix-gate:
     name: arduino-matrix-gate
-    runs-on: ubuntu-latest
+    runs-on: ${{ fromJSON(vars.CI_RUNNER || '["ubuntu-latest"]') }}
     timeout-minutes: 5
     needs: arduino-matrix
     if: always()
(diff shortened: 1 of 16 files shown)


w1ne:
Closing — the premise was wrong.

I justified this with "when hosted runners went unavailable on 2026-08-02, this repo's CI had no way to move." That never applied here: **labwired-core is public, so its GitHub-hosted minutes are free and unmetered.** The billing block only affects the private repos (`labwired`, `labwired-landing`). This repo's CI ran green straight through the outage — at 19:06 and 20:07 on hosted runners — which I mistakenly read as evidence the block had cleared.

So a kill-switch here buys nothing and adds indirection to 31 job definitions. Not worth carrying.

The private repos already route through `vars.CI_RUNNER` to self-hosted, which is the protection that actually matters.

PR #757, opened 2026-08-03
feat(inspect): one seam for device binding, one for display evidence

w1ne:
## One seam for binding, one seam for evidence

Two defects with the same shape: a device could be perfectly simulated and still be
invisible, because the code that reported it was a hand-maintained list somewhere else.

### 1. Binding was not universal

A placed part became attached in one of two structurally different ways, and only one
was visible to `inspect`:

* **Owned by a controller** — an I²C slave or SPI panel lives inside the controller it
  answers to, reachable only via `Peripheral::for_each_attached_device`.
* **Resident on the bus** — an HC-SR04 on TRIG/ECHO, a TM1637 clocked by the GPIO
  write-hook, a servo on a PWM pad, a WS2812 strip, a thermistor driving an ADC channel,
  a CAN tester node. None answers to an address, so each lived in its own typed
  collection on `SystemBus` — **fourteen** of them, bound through seven ad-hoc paths —
  and none of it reached `inspect`.

Now one walk (`SystemBus::for_each_attached_device`), one borrowed record
(`AttachedDeviceRef`), one join (`SystemBus::inspect_devices`, moved off `Machine`
because the bus is the one thing holding BOTH the live models and the manifest
declarations). **A transport is a field of the record, not a branch in the pipeline.**

Controller-hosted devices are walked first and bus-resident ones after, so the addressed
join — which does real disambiguation — is never disturbed by a name-matched device, and
the existing `devices` array stays a stable prefix of what it was.

Two shape changes, both to avoid reporting a value the model does not have:

* `DeviceAttachment.bus` is now `Option<String>` (`skip_serializing_if`). Every device
  that came from a manifest has an owner, so the key is present on every real rig; a
  programmatic attachment with nothing to name its connection reports unknown rather
  than a plausible guess. JSON for the existing I²C/SPI path is byte-identical.
* `channel` widens from "bus-switch channel" to "which channel of `bus`", covering an
  ADC channel too.

### 2. There were TWO display-evidence systems, and neither knew about the other

`inspect::device_artifacts` was a central `match` on concrete types with exactly two
arms — `Ssd1306` and `Ili9341` — in a file far from any model. Meanwhile
`crates/wasm/src/inspect.rs` carries a hand-written per-panel × per-controller downcast
matrix of `get_*_framebuffer` accessors that renders **ten** panels.

So the SH1107 painted a full console in the browser while `inspect` reported no artifact
at all. Both were telling the truth about their own path.

That was not cosmetic. `labwired_verify`'s display oracle resolves `painted` /
`min_ink_bytes` / `min_refresh_generation` against these artifacts, so a lab built on one
of those panels **could never pass a display clause however correct its firmware** — and
the tool documents `panel: "ssd1680_tricolor_290"` as an example value the engine could
not emit.

Adding three more arms would have fixed today's three panels and rebuilt the trap for the
fourth. Instead a model reports its own evidence by overriding `artifacts` on the device
trait it already implements, next to the buffers it reads. `device_artifacts` is gone;
`AttachedDeviceRef` carries `Option<&dyn DeviceEvidence>` instead of `Option<&dyn Any>`,
and the join calls it once it has resolved the id an artifact is addressed by.

Because binding is now universal, a display wired to **pins** can report too: the TM1637,
the direct-driven 7-segment digit and the WS2812 strip implement the evidence seam
directly and ride the same walk. None of them could have, before — **this is why the two
fixes belong in one change.**

**Eleven models** report: SSD1306, SH1107, ILI9341, PCD8544, SSD1680 tri-colour, UC8151D
tri-colour, MAX7219, LCD1602, TM1637, seven-segment, WS2812.

The e-paper representation is deliberate: ink is a byte that is **not** `0xFF` (the
panel's own no-ink convention, and the same count the CLI's `black-plane non-FF bytes=`
line prints), the two planes are reported separately with a `plane_bytes` split rather
than a synthesized composite, and `refresh_generation` is surfaced — which is what
finally makes `min_refresh_generation` reachable. The LCD1602, TM1637 and 7-segment
report `kind: "text_display"`, not a fake framebuffer.

The bus-trace decorators forward `artifacts`. A decorator that let it fall through to the
trait default would silently erase the wrapped panel's evidence — it did, on the first
run, and the existing ILI9341 test caught it.

**No fidelity is added.** Every number is read off the model's own buffer; a panel whose
driver never painted reports zero, and a device with no evidence returns an empty list.
**SSD1306 and ILI9341 payloads moved verbatim** — same keys, same definitions, same
counts.

---

### Every test was derived from outside the implementation and watched to fail first

`crates/core/tests/inspect_device_binding_universal.rs` reads the shipped
`examples/*/system.yaml` files with a YAML parser and asserts every placed device appears
in `inspect`. On the unchanged tree:

```
test bay_occupancy_i2c_and_spi_still_inspectable ... ok
test analog_thermistor_is_inspectable ... FAILED
test tm1637_bit_banged_display_is_inspectable ... FAILED
test can_log_player_node_is_inspectable ... FAILED
test hc_sr04_on_gpio_pins_is_inspectable ... FAILED
test can_uds_tester_node_is_inspectable ... FAILED

examples/ntc-thermistor-lab/system.yaml: device 'thermistor' (ntc-thermistor) is placed
on 'adc1' but does not appear in inspect's devices array; inspect reported []

examples/nokia5110-invaders-lab/system.yaml: device 'dist' (hc-sr04) is placed on
'gpioa' but does not appear in inspect's devices array; inspect reported ["lcd"]

examples/h563-uds-ecu/system.yaml: device 'uds-tester' (uds-tester) is placed on
'fdcan1' but does not appear in inspect's devices array; inspect reported []

test result: FAILED. 1 passed; 5 failed; 1 ignored
```

`crates/core/tests/panel_artifact_evidence.rs` drives each panel over its **own wire
protocol** and checks reported counts against arithmetic on the test's own input. On the
unchanged tree:

```
test ssd1306_evidence_is_unchanged ... ok
test ili9341_evidence_is_unchanged ... ok
test sh1107_that_painted_reports_its_ink ... FAILED
test ssd1680_that_refreshed_reports_both_planes_and_the_refresh ... FAILED
test uc8151d_that_refreshed_reports_both_planes_and_the_refresh ... FAILED
test pcd8544_that_painted_reports_its_ink ... FAILED
test max7219_that_was_written_reports_its_lit_leds ... FAILED
test lcd1602_that_was_written_reports_its_text ... FAILED
test every_panel_the_browser_renders_reports_evidence_to_inspect ... FAILED

panel 'panel' (Some("ssd1680_tricolor_290")) produced NO artifact; the display oracle
has nothing to resolve against, so no lab using this panel can be verified however
correct its firmware is

the browser renders 'ssd1306' but its model reports no artifacts to inspect — every
oracle clause about that panel is unresolvable.

test result: FAILED. 2 passed; 9 failed
```

The two that passed are exactly the two panels that already had an arm.

### The existing device list is byte-identical

`inspect`'s `devices` array for the bay-occupancy rig — the customer delivery, the one
with the TCA9548A, four VCNL4010s and the ILI9341 with its framebuffer artifact — was
captured before and after and diffed:

```
bay-occupancy devices: BYTE-IDENTICAL
```

Everything else in the diff is pure addition, e.g. the Nokia-5110 rig which reported one
device and now reports two:

```
   { "id": "lcd", "device_type": "pcd8544", ... ,
+    "artifacts": [ { "kind": "framebuffer", "meta": { "format": "pcd8544_bank", ... } } ]
+  },
+  { "id": "dist", "device_type": "hc-sr04", "declared": true,
+    "attachment": { "transport": "gpio", "bus": "gpioa" }, "artifacts": [] }
```

### Full suite vs a pristine `origin/main` baseline

Both `cargo test --workspace --no-fail-fast`, same machine.

| | pristine baseline | this branch |
|---|---|---|
| failing targets | 8 | 5 |

Failing on this branch — **all five also fail on pristine `origin/main`**:
`esp32_ili9341_attach`, `register_coverage`, `world_can_bus`,
`world_esp32c3_pingpong`, `labwired-dap --test e2e`.

The three that no longer fail (`demo_blinky`, `strict_onboarding`,
`e2e_stimulus_reporting`) were failing on missing pre-built firmware artifacts, which got
built during the session. **No new failure.**

`cargo clippy -D warnings` is clean for `labwired-core`, `labwired-cli` and
`labwired-dap`. Two pre-existing lint failures on this toolchain in files this branch
does not touch (`crates/firmware-nrf52840-proximity/src/main.rs` manual-checked-ops,
`crates/wasm/src/lib.rs` manual-div-ceil) are unchanged.

### Ryan's bay-occupancy rig: 12/12 PASS

```
PASS  1,2,3   init, per-read channel select, independent injection
PASS  4       raw counts -> EMPTY/PRESENT via thresholds
PASS  5       separate entry/exit thresholds (hysteresis)
PASS  6       debounce: single noisy spike rejected
PASS  7       simultaneous occupancy across all four bays
PASS  8       one channel cannot change another's state
PASS  9       missing sensor / I2C failure detected
PASS  10      per-bay state painted on the TFT  (panel ink=6610)
PASS  11      clear FAULT state for an unreadable sensor  (panel ink=6610)
PASS  12      polling and display do not block each other  (panel ink=6352)
```

### Ratchets, so neither defect can come back quietly

* `attached_device_walk_covers_every_bus_collection` parses the `SystemBus` declaration
  and fails if a collection of device models is not walked. A fifteenth collection added
  and forgotten is a build failure.
* `every_panel_the_browser_renders_reports_evidence_to_inspect` derives its list from the
  wasm renderer's own `get_*_framebuffer` accessors — the *other* system — so the two
  cannot drift apart again. It is what surfaced the TM1637 and 7-segment gaps.
* `record_external_devices_has_one_home` (pre-existing) still holds.

---

## What is deliberately NOT in this PR

The wasm `get_*_framebuffer` accessors still re-derive raw framebuffers through their own
per-panel × per-controller downcast matrix. This PR does not extend that matrix and does
not touch it. Converting those accessors onto `for_each_attached_device` +
`DeviceEvidence` is a follow-up, because it needs browser verification that every panel
which renders today still renders — a claim a Rust test suite cannot support, and a
regression there would be worse than the bug being fixed here.

After this PR there is exactly ONE place that decides **what a panel reports as
evidence** (the model itself); the renderer's remaining job is fetching raw pixels.

`Category` in the kit registry distinguishes transports, not displays, so the panel
ratchet derives its list from the wasm renderer rather than from the registry. Giving the
registry a display category is the cleaner long-term source and is worth doing when the
wasm conversion lands.

### DIFF
--- a/crates/core/src/bus/attached_devices.rs
+++ b/crates/core/src/bus/attached_devices.rs
@@ -171,14 +171,16 @@ impl SystemBus {
             self.emit_resident(f, Resident::gpio(dev.id(), None));
         }
         for dev in &self.tm1637 {
-            self.emit_resident(f, Resident::gpio(&dev.id, None));
+            // A bus-resident DISPLAY: it reports evidence directly, because it
+            // has no controller trait to hang it on.
+            self.emit_resident(f, Resident::gpio(&dev.id, Some(dev)));
         }
         for dev in &self.hx711 {
             let id = dev.component_id().unwrap_or("hx711");
             self.emit_resident(f, Resident::gpio(id, None));
         }
         for dev in &self.seven_segment {
-            self.emit_resident(f, Resident::gpio(&dev.id, None));
+            self.emit_resident(f, Resident::gpio(&dev.id, Some(dev)));
         }
         for dev in &self.analog_inputs {
             // An analog source has no `Any` view, so it is listed but not read:
@@ -363,9 +365,11 @@ struct Resident<'a> {
     bus: Option<&'a str>,
     channel: Option<u8>,
     /// A bus-resident model that can show something of itself implements
-    /// [`crate::inspect::DeviceEvidence`] directly and is passed here. None of
-    /// them does today — a servo and a distance sensor have no display surface
-    /// — and `None` says exactly that rather than promising an empty screen.
+    /// [`crate::inspect::DeviceEvidence`] directly and is passed here — the
+    /// TM1637 and the direct-driven 7-segment digit are displays that happen to
+    /// be wired to pins. `None` is the honest answer for everything else: a
+    /// servo and a distance sensor have no display surface, and `None` says
+    /// that rather than promising an empty screen.
     evidence: Option<&'a dyn DeviceEvidence>,
 }
 
--- a/crates/core/src/bus/attached_devices.rs
+++ b/crates/core/src/bus/attached_devices.rs
@@ -147,7 +147,7 @@ impl SystemBus {
         }
         for dev in &self.ws2812 {
             let id = dev.component_id().unwrap_or("ws2812");
-            self.emit_resident(f, Resident::gpio(id, None));
+            self.emit_resident(f, Resident::gpio(id, Some(&**dev)));
         }
         for dev in &self.servos {
             self.emit_resident(f, Resident::gpio(dev.id(), None));
--- a/crates/core/src/bus/attached_devices.rs
+++ b/crates/core/src/bus/attached_devices.rs
@@ -1,0 +1,416 @@
+// LabWired - Firmware Simulation Platform
+// Copyright (C) 2026 Andrii Shylenko
+// SPDX-License-Identifier: MIT
+
+//! ONE walk over every device an author placed on this machine, whatever
+//! transport binds it, and the ONE join that gives each one its manifest
+//! identity.
+//!
+//! # Why this is not per-transport
+//!
+//! A placed part becomes attached to the machine in one of two structurally
+//! different ways, and for a while only the first was visible to `inspect`:
+//!
+//! 1. **Owned by a controller.** An I²C slave or an SPI panel lives inside the
+//!    controller peripheral it answers to, so it is reachable only by asking
+//!    that controller — [`crate::Peripheral::for_each_attached_device`]. A walk
+//!    over `SystemBus::peripherals` alone cannot see it. That is the bug that
+//!    made a customer rig report 52 chip-internal peripherals and not one of
+//!    the six parts on the canvas.
+//! 2. **Resident on the bus.** An HC-SR04 on TRIG/ECHO, a servo on a PWM pad, a
+//!    WS2812 strip, a thermistor driving an ADC channel, a CAN tester node —
+//!    none of these answers to an address, so none can live inside a
+//!    controller. Each is held in a typed collection on [`SystemBus`] instead.
+//!
+//! The second kind was invisible to `inspect` for exactly the same reason the
+//! first kind had been: the walk did not go where the models are. Fixing it per
+//! family would have meant an eighth, ninth, tenth special case — each one a
+//! fresh chance to add a device that simulates correctly and reports nothing.
+//!
+//! So there is one walk ([`SystemBus::for_each_attached_device`]) that covers
+//! both kinds, one borrowed record shape
+//! ([`crate::inspect::AttachedDeviceRef`]) that both kinds emit, and one join
+//! ([`SystemBus::inspect_devices`]) that turns either into the same
+//! [`crate::inspect::DeviceInspect`]. A transport is a field of the record, not
+//! a branch in the pipeline.
+//!
+//! # What the two kinds know about themselves
+//!
+//! They differ in exactly one respect, and the record shape carries it
+//! explicitly rather than hiding it:
+//!
+//! * A controller-hosted device is identified by WHERE it answers. It states
+//!   its address (or chip-select) and the controller states the bus; the
+//!   manifest name is joined on afterwards by matching that placement.
+//! * A bus-resident device has no address to be identified by, so it carries
+//!   the `external_devices:` id it was stamped with when it was attached, and
+//!   the join matches on that. This is not a fallback or a guess: the id is the
+//!   author's own text, copied onto the model at attach.
+//!
+//! # The order is load-bearing
+//!
+//! Controller-hosted devices are walked first, bus-resident ones after. A
+//! declaration can only be claimed once, so this guarantees the addressed join
+//! — which has to do real disambiguation work — is never disturbed by a
+//! name-matched device, and keeps the `devices` array a stable prefix of what
+//! it was before bus-resident devices joined it.
+
+use super::SystemBus;
+use crate::inspect::{AttachedDeviceRef, DeviceEvidence};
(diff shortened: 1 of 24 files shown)


w1ne:
### Bay-occupancy rig: 12/12 PASS on this branch

Run from `examples/esp32-bay-occupancy/tests` against this branch's release CLI:

```
PASS  1,2,3   init, per-read channel select, independent injection
PASS  4       raw counts -> EMPTY/PRESENT via thresholds
PASS  5       separate entry/exit thresholds (hysteresis)
PASS  6       debounce: single noisy spike rejected
PASS  7       simultaneous occupancy across all four bays
PASS  8       one channel cannot change another's state
PASS  9       missing sensor / I2C failure detected
PASS  10      per-bay state painted on the TFT  (panel ink=6610)
PASS  11      clear FAULT state for an unreadable sensor  (panel ink=6610)
PASS  12      polling and display do not block each other  (panel ink=6352)
```

The ink counts matter more than the PASS marks: **6610 / 6610 / 6352 are identical to the values measured on `origin/main` before this work**. The ILI9341 artifact payload is unchanged byte for byte, which is the specific guarantee this PR owes — a panel that already reported correctly must keep reporting exactly the same thing.

Full `cargo test -p labwired-core` still running; will report it here as a fact before this is merged.

w1ne:
### Full suite: 2614 passed, 1 failed, 0 compile errors

The single failure is `a_device_type_the_factory_cannot_build_is_an_error_not_a_warning` (`crates/core/tests/esp32_ili9341_attach.rs:74`).

That test is about **device attachment**, which is precisely what this PR rewrites, so I did not take "known red" on trust. Checked out pristine `origin/main` (`6d0e2a23`) in a separate worktree and ran it there:

```
thread 'a_device_type_the_factory_cannot_build_is_an_error_not_a_warning' panicked at
crates/core/tests/esp32_ili9341_attach.rs:74:10:
an unbuildable device type must fail the attach, not be skipped: ()

test result: FAILED. 1 passed; 1 failed; 0 ignored
```

Identical failure, identical line, on a tree that contains none of this work. **Pre-existing, not caused here.**

So this branch is: **2614 passed / 1 inherited failure / 0 compile errors**, plus bay-occupancy 12/12 at ink 6610 / 6610 / 6352 (byte-identical to main). Ready to merge on that evidence.

PR #766, opened 2026-08-03
ci: CI fixes formatting instead of failing on it

w1ne:
`cargo fmt --all -- --check` was a step in `pr-gate` — a REQUIRED check. It failed three times in one evening on a change whose content was already correct: the machine printed the exact diff it wanted and then refused to apply it. Whitespace is the one class of review comment a machine can settle by itself, so now it does.

## What changed

**New `fmt-autofix` job** (`pull_request` only). Checks out the PR **branch tip** (not the merge commit — the fixup has to land somewhere), installs the **pinned 1.95.0** toolchain, runs `cargo fmt --all`, and pushes the result back to the branch if anything moved.

**`pr-gate` loses its `Check Formatting` step.** Nothing else in `pr-gate` is touched.

**`core-integrity` keeps its `Check Formatting` step**, now with a comment saying it is the backstop. That is the answer to "is formatting still enforced anywhere": yes, on push to main, where it *reports* instead of gating a merge.

## Token: a PAT, not `GITHUB_TOKEN`

`GITHUB_TOKEN` + `permissions: contents: write` **can** push. It is still the wrong choice here, and not for the usual "it wastes a CI run" reason:

- `main` requires exactly two contexts — **`pr-gate`** and **`release-runner-contract`** — with `enforce_admins: true`.
- Both are `pull_request`-only. This workflow's `push` trigger is `branches: [main]`, and `pr-gate` is additionally gated on `github.event_name == 'pull_request'`.
- Required contexts are evaluated against the PR's **head SHA**.
- A `GITHUB_TOKEN` push deliberately fires **no** workflow run.

So a `GITHUB_TOKEN` fixup advances the head SHA past both required contexts, nothing re-reports them, and the PR sits **BLOCKED** — "Expected — waiting for status to be reported" — with every check it *did* run green, unfixable without a human pushing another commit. That is strictly worse than the red `fmt --check` this replaces. It is the same failure mode that stranded three superproject PRs on 2026-08-02, and it is why `submodule-bump.yml` refuses the same fallback.

A PAT push is an ordinary user push: `pull_request` fires, both required contexts land on the new head, the PR stays mergeable.

⚠️ **Action required to enable this:** set repo secret **`CI_FMT_PAT`** on `w1ne/labwired-core` — a fine-grained PAT with **`contents: write`** on this repo and nothing else. Until then the job green-skips with a message.

## Never red

Two green-skip paths, both printing a `::notice::` that says what to do:
- **fork PR** — read-only token, secrets withheld, push impossible by construction;
- **no `CI_FMT_PAT`** — as above.

A formatting job that fails is exactly what is being removed, so it does not fail for either.

## Loop guard

The PAT push fires a fresh run, which is the *point* (it re-reports the required checks). It terminates because rustfmt is idempotent: the re-run finds no diff and pushes nothing. At most **one** fixup per human push.

The explicit flag covers the case where that is not true — if the tip commit is already a `style: cargo fmt --all [ci-autofix]` fixup **and** `cargo fmt` still wants changes, formatting is not converging and the job stops loudly instead of starting a cycle. That is the one condition under which it goes red, and it should.

No `[skip ci]` in the fixup message: that would suppress the very run that re-reports `pr-gate`.

## Pinned toolchain

`dtolnay/rust-toolchain@1.95.0`, matching `pr-gate` and `core-integrity`. Local dev here is on 1.96.1 — the same drift that made `clippy::redundant_closure` fire only in CI would, unpinned, mean CI reformats what a developer's local `cargo fmt` just formatted, on every push, forever. The job logs `cargo fmt --version` so this is checkable from the run, not just from the YAML.

## Proof

Commit 2 on this branch deliberately misformats `crates/core/src/sched/result.rs` — squashed derive commas, an eight-space indent, padded colon spacing. Valid Rust, clippy-clean, and exactly what used to fail `pr-gate`. If this works, CI's own commit restores that file byte-for-byte to `origin/main` and the net diff of this PR is workflow YAML only.

`actionlint` clean.


w1ne:
## Proof, from real CI

**Run 1 — [30849423280](https://github.com/w1ne/labwired-core/actions/runs/30849423280) (head `885701c`, the deliberately misformatted commit)**

```
rustfmt 1.9.0-stable (59807616e1 2026-04-14)
rustc 1.95.0 (59807616e 2026-04-14)          <- PINNED. local dev here is 1.96.1
rustfmt wants these changes:
 crates/core/src/sched/result.rs | 6 +++---
[ci/auto-format-pr c21f251] style: cargo fmt --all [ci-autofix]
To https://github.com/w1ne/labwired-core
   885701c..c21f251  HEAD -> ci/auto-format-pr
Pushed the formatting fixup to ci/auto-format-pr.
```

**The branch converged, byte-for-byte.** `crates/core/src/sched/result.rs` blob at the merge base and at the branch tip are the same object:

```
$ git rev-parse 97ddb7c6:crates/core/src/sched/result.rs \
                 ci/auto-format-pr:crates/core/src/sched/result.rs
64a1bd50f008e00367abc2132c381fee92830b52
64a1bd50f008e00367abc2132c381fee92830b52
```

Net content diff of this PR vs the merge base is **workflow YAML only** — the misformatting commit exists solely to be undone by CI, and it was.

---

### The PAT decision, confirmed by the run rather than argued

The fixup push produced **run [30849464093](https://github.com/w1ne/labwired-core/actions/runs/30849464093), `pull_request`, head `c21f251`**. That run is the whole reason for the PAT: it is what re-reported `pr-gate` and `release-runner-contract` against the new head SHA. Pushed with `GITHUB_TOKEN` that run would not exist, `c21f251` would carry zero required contexts, and this PR would read `BLOCKED` with everything green.

### `pr-gate` green with no formatting step

Run 2, `pr-gate` — every step, in order:

```
1  success  Set up job
2  success  Run actions/checkout@v4
3  success  Install Rust
4  success  Cache dependencies
5  success  Set up Python
6  success  Validation status — regen check + drift gate
7  success  Generated artifacts — regenerate and assert no drift
8  success  Clippy (default-members)
9  success  Unit tests (default-members lib)
10 success  Walk-deletion derivation (PR-fast structural arm)
```

No `Check Formatting`. Nothing else moved.

### Loop guard holds

`fmt-autofix` on run 2, i.e. the run its own push created:

```
Tip commit is a previous autofix — this run must not push again.
Formatting converged: the previous autofix is stable under a re-run.
```

Nothing pushed. **Exactly two runs exist on this branch** and the tip has not moved off `c21f251`:

```
30849464093 pull_request c21f251 completed/success
30849423280 pull_request 885701c completed/cancelled   <- superseded by cancel-in-progress
```

### Green-skip with no PAT

The proof secret has been **deleted** (see below), and `fmt-autofix` was re-run in that state — job **success**, steps 3-5 `skipped`, with this annotation:

> **notice: fmt-autofix skipped** — Secret CI_FMT_PAT is not set. GITHUB_TOKEN is not a fallback: its pushes fire no workflow run, so the new head SHA would carry neither required context (pr-gate, release-runner-contract) and the PR would be unmergeable. Add a PAT with contents:write to enable auto-formatting.

---

### ⚠️ Before this does anything: set `CI_FMT_PAT`

The evidence above was produced with a **temporary** `CI_FMT_PAT` that has since been **deleted** — `w1ne/labwired-core` currently has **zero** Actions secrets, so `fmt-autofix` green-skips on every PR until one is added.

It was deleted on purpose. The only credential available to run the proof was the local `gh` OAuth session token (`gist, read:org, repo, workflow`), which reaches **every** repo on the account including private ones, and it rotates on re-auth. That does not belong in a public repo's Actions secrets as a standing credential.

Mint a **fine-grained PAT** scoped to `w1ne/labwired-core` only, permission **Contents: Read and write**, nothing else, and:

```
gh secret set CI_FMT_PAT --repo w1ne/labwired-core
```

### Also note

This PR is `BEHIND` because #760 merged while the proof was running — `main` requires branches be up to date. Unrelated to this change; merge `main` in (never rebase) before merging.

PR #767, opened 2026-08-03
One bus trace for every wired bus (and the 4-of-5 UARTs it un-silenced)

w1ne:
## The defect, measured

`bus/bus_trace.rs` calls itself the *universal* bus-transaction trace. It was universal over two buses.

| Bus | Ring | Limit | Seq | Cycle |
|---|---|---|---|---|
| I²C / SPI | shared `BusTrace` | 1024 | shared | ✅ |
| UART | per-instance `VecDeque` | 512 | own | ❌ |
| CAN | per-instance `VecDeque` | 200 | own | ❌ |

`grep -c bus_trace` was **0** in `uart.rs`, `fdcan.rs`, `bxcan.rs`, `can.rs`.

That fork was already broken, three ways:

1. **Four of five UART models traced nothing at all.** `wasm/traces.rs` located UARTs with `downcast_ref::<Uart>()`. `EspUart` (C3/S3), `Esp32Uart`, `Nrf52Uarte` and `Nrf54lUarte` are all UARTs and none of them **is** a `Uart` — so on every ESP and nRF lab the analyzer's UART panel was permanently empty, silently, because *"no UART found"* and *"a UART that sent nothing"* are the same empty array. This is the identical mistake `UartStreamHost` was introduced to fix for cross-chip links; the trace path never got the same treatment.
2. **No cycle stamp on UART or CAN**, so `BusTraceEvent::cycle`'s documented promise to "time-align protocol decode with sampled waveforms on one cycle axis" held for only half the buses.
3. **Three sequence counters** ⇒ cross-bus ordering unrecoverable. Nothing could answer "did this UART byte precede that I²C address phase?"

## One ring, one envelope, one door

- `BusPayload` gains `Uart { direction, byte }` and `Can { .. }`. The **envelope** is universal (one seq, one cycle, one bus name); the **payload** stays honest about the protocol — an I²C symbol genuinely is not a CAN frame.
- `Peripheral::attach_bus_trace` joins `attach_cycle_clock` / `attach_irq_line` at the three registration funnels in `bus/construct.rs`. A model reaches the trace by *being registered*, not via a per-family call site. I²C/SPI keep the wrapper funnel — structurally unbypassable, no hook needed.
- All seven models record into the shared ring. Private rings, `UART_TRACE_LIMIT` and both `trace_seq` counters deleted. The per-model accessors survive as **derived projections**, which is why the pre-existing UART/CAN trace tests pass untouched — they are the byte-identity proof, written before this refactor.
- The bridge stops downcasting. A new UART or CAN family now joins by **recording**, not by someone remembering to add a `downcast_ref` arm.

## Ring capacity: 1024 → 4096

Each bus used to have its own budget, so a chatty UART could not evict an I²C address phase. With one ring it can, and eviction is silent — an instrument reconstructing transactions from address phases just sees fewer, not an error. `firmware_survival.rs:1741` already documents this exact hazard for SPI redraws vs I²C. ≈330 KB worst case with 64-byte CAN-FD payloads.

## Guards — each watched failing before it was kept

- **`every_recording_peripheral_shares_the_machines_ring`** — fails **closed**. For every chip descriptor on disk, a peripheral that admits to recording must hold the bus's ring by `Arc` identity. An un-injected model keeps its private handle and writes to a buffer nobody reads; comparing *contents* cannot detect that (two empty rings are equal), comparing *identity* can. Deliberate break (unwiring one funnel) fails it across `esp32`, `esp32c3`, `esp32s3`, `mkw41z4`, `nrf52832`, `nrf52840`.
- **`no_peripheral_keeps_a_private_trace_ring`** — fails if **forked**. A source scan, because a private ring is invisible at runtime by construction: the model behaves correctly, tests green, and is simply absent from the trace. Allowlist carries the air traces (different medium, already one home behind `VirtualAirBus`) and IO-Link (a decode of UART octets now in the ring), each with its reason.
- **`the_private_ring_detector_is_not_vacuous`** pins the detector against the three declarations this PR removes, so the scan cannot rot into always-green.
- **`an_esp_uart_reaches_the_shared_trace`** — the end-to-end property that was broken: a C3 UART0 byte must appear in `bus_trace_snapshot()`.

## Scope

Air traces (BLE / WiFi / radio) are deliberately **out**: an over-the-air frame has no wire and no bus name, and they already share their own one home. Wired and air are two honest axes; collapsing them would make the model worse. IO-Link is out for now and flagged as debt in the allowlist — it is a decode layer above UART octets, which are now in the ring, so it is derivable.

## Verification

- **2612 tests across 30 targets, 0 failures** (`cargo test -p labwired-core`, all targets incl. `firmware_survival`).
- `cargo clippy` clean for both `labwired-core` and `labwired-wasm`. Note `crates/wasm` is not a workspace default-member, so it was built explicitly for `wasm32-unknown-unknown` — the core PR lane never links it.

Follow-up (separate PR, monorepo): bump the submodule pointer and collapse `traceSnapshot(kind)` plus the four per-protocol converters in `wireActivity.ts` into one, now that the engine speaks one dialect.

### DIFF
--- a/crates/cli/src/bus_vcd.rs
+++ b/crates/cli/src/bus_vcd.rs
@@ -13,7 +13,9 @@
 //!   bus, so the capture opens directly in GTKWave, PulseView/sigrok, or any
 //!   other waveform viewer. Each event becomes one value-change at `#<seq>`
 //!   carrying the transacted byte (I²C: the `byte` field, already covering
-//!   the address frame via `I2cSym::AddrWrite`/`AddrRead`; SPI: `mosi`).
+//!   the address frame via `I2cSym::AddrWrite`/`AddrRead`; SPI: `mosi`; UART:
+//!   the octet). CAN is frame-oriented and has no 8-bit wire sample, so it
+//!   appears in the JSON export but not the VCD — see `event_byte`.
 
 use labwired_core::bus::bus_trace::{BusPayload, BusTraceEvent};
 use std::collections::BTreeMap;
@@ -25,11 +27,23 @@ pub fn write_bus_trace_json<W: Write>(events: &[BusTraceEvent], w: W) -> io::Res
     serde_json::to_writer_pretty(w, events).map_err(io::Error::other)
 }
 
-/// The byte a `BusTraceEvent` carries on the wire, for the VCD vector signal.
-fn event_byte(payload: &BusPayload) -> u8 {
+/// The byte a `BusTraceEvent` carries on the wire, for the VCD vector signal —
+/// or `None` when the event is not a single-byte wire symbol at all.
+///
+/// CAN is the `None` case, and deliberately so. A CAN event is a whole FRAME
+/// (arbitration id, DLC, up to 64 payload bytes); there is no one octet that
+/// represents it on an 8-bit vector. Emitting its first data byte, or a zero,
+/// would put a value in a waveform viewer that never existed on the wire —
+/// worse than omitting it, because a waveform is read as measurement. A
+/// frame-oriented bus wants its own VCD representation, which is a separate
+/// piece of work; until then the JSON export carries CAN losslessly and the VCD
+/// says nothing rather than something false.
+fn event_byte(payload: &BusPayload) -> Option<u8> {
     match payload {
-        BusPayload::I2c { byte, .. } => *byte,
-        BusPayload::Spi { mosi, .. } => *mosi,
+        BusPayload::I2c { byte, .. } => Some(*byte),
+        BusPayload::Spi { mosi, .. } => Some(*mosi),
+        BusPayload::Uart { byte, .. } => Some(*byte),
+        BusPayload::Can { .. } => None,
     }
 }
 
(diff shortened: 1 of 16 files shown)


w1ne:
## Verification update (after `ba5a6c60`)

The first push was red. Two real defects, both reachable only by a wider build than I had been running (`-p labwired-core --lib` reaches neither):

1. **`machine_advance` lost its `#[cfg(test)]`.** Adding `bus_trace_one_home` to `tests/mod.rs` by inserting *before* `pub mod machine_advance;` handed my module that module's attribute and left `machine_advance` ungated. It then compiled into the plain lib target where its test-only helpers are dead code — 10 warnings, fatal under `-D warnings`. Locally these surfaced only as warnings, so `cargo test` and a plain `cargo clippy` both stayed green.
2. **`crates/cli/src/bus_vcd.rs` matched `BusPayload` exhaustively** (`E0004`). Two new variants broke it.

`cargo clippy --all-targets -- -D warnings` — CI's exact command — now exits 0.

### The wasm bridge is now actually verified

`crates/wasm` is not a workspace default-member, so **no check on this PR links it** — the `uart_trace_snapshot` rewrite, i.e. the part that fixes the user-visible bug, had no gate. Built the bundle and drove a real ESP32-C3 lab (`esp32c3-devkit.yaml` + `esp32c3-demo.elf`) through it:

```
uart_trace_snapshot() entries : 1
  peripheral=uart0 events=7  tx:0x45@cyc13975  tx:0x53@cyc27854  tx:0x50@cyc41734  tx:0x20@cyc55614
bus_trace_snapshot() by proto : {"uart":7}

PASS  uart_trace_snapshot() is non-empty for a C3 lab (was [] before: EspUart is not a Uart)
PASS  the shared ring carries protocol="uart" rows (BusPayload::Uart through serde_wasm_bindgen)
PASS  direction serializes as "tx"/"rx" — byte-identical to the &'static str it replaced
PASS  UART events now carry a cycle stamp (they never could before)
```

`0x45 0x53 0x50 0x20` is `"ESP "` — the firmware's console output, visible to the analyzer for the first time.

### One flake, not a regression

`e2e_kw41z_cow_stimulus` failed once with `parse result.json: expected ',' or '}' at line 616`. I initially read "passes on main, fails here" as causation. It is not: `result.json` contains no trace data at all (no `bus_trace`, no `protocol`), and line 616 of a 14,492-line file is a **truncated** write, not corrupted content. Re-ran on this branch **6/6 green** (3 serial, 3 parallel) at 37s vs main's 38.8s — no slowdown. The one failure coincided with two other heavy cargo builds on the same machine; the test derives its output dir from `SystemTime::now().as_nanos()` and runs two cases in parallel. Pre-existing contention flake, left alone rather than folded into this PR.

PR #827, opened 2026-08-06
fix(perf): real baselines, all 25 chips covered, and one issue instead of a comment per push

### LINKED ISSUE #778: Simulator throughput regression (per-board perf gate)
Commit: eae0786b6224b92e078f684d3a1c60fa98e8d1b7
Run: https://github.com/w1ne/labwired-core/actions/runs/31005780275

```

simulator throughput regressed:
  stm32f103: 742.0 -> 1438.0 Ir/step (+93.8%)
  stm32f401: 742.0 -> 1230.0 Ir/step (+65.8%)
  stm32f401cdu6: 742.0 -> 1662.0 Ir/step (+124.0%)
  stm32f407: 742.0 -> 1494.0 Ir/step (+101.3%)
  stm32g474re: 742.0 -> 1220.2 Ir/step (+64.4%)
  stm32l476: 742.0 -> 1884.2 Ir/step (+153.9%)
  stm32wb55: 742.0 -> 1206.0 Ir/step (+62.5%)

Every extra host instruction per simulated step slows the browser twin by the same proportion.
If the cost is intentional, re-baseline with: python3 scripts/perf/board_perf.py --update
board               Ir/step   baseline     delta
-------------------------------------------------
stm32f103            1438.0      742.0   +93.8%  REGRESSION
stm32f401            1230.0      742.0   +65.8%  REGRESSION
stm32f401cdu6        1662.0      742.0  +124.0%  REGRESSION
stm32f407            1494.0      742.0  +101.3%  REGRESSION
stm32g474re          1220.2      742.0   +64.4%  REGRESSION
stm32h563            1605.5     3634.5   -55.8%  (faster)
stm32h735            1378.5     2729.5   -49.5%  (faster)
stm32l073            1596.5     3390.5   -52.9%  (faster)
stm32l476            1884.2      742.0  +153.9%  REGRESSION
stm32wb55            1206.0      742.0   +62.5%  REGRESSION
stm32wba52           1180.5     2380.5   -50.4%  (faster)

not covered by this gate:
  nrf52840 / nrf52832 / nrf5340 / nrf54l15: flash @ 0x00000000 — needs its own fixture link script
  rp2040: flash @ 0x10000000 — needs its own fixture link script
  mkw41z4: flash @ 0x00000000 — needs its own fixture link script
  esp32 / esp32s3 / esp32c3: Xtensa / RISC-V — needs a per-ISA fixture

```

Re-baseline with `python3 scripts/perf/board_perf.py --update` if the cost is intentional.

w1ne:
## Description

The per-board perf gate had been filing a 60–150% "REGRESSION" against seven STM32 boards on **every push to `main`** for a day, as a fresh comment on #778 each time. None of it was real, and the gate was measuring 11 of 25 chips.

### The regressions were never real

`scripts/perf/baselines.json` was committed with `"742.0"` as the baseline for **seven different boards at once**. Seven chips with different peripheral counts cannot share a per-step cost — the real spread across those same boards is 1261–1939 Ir/step. The number was typed, not measured, so every run since compared a real measurement against a placeholder.

The other four boards had the opposite problem: baselines roughly **2× above** what they measure. That is not a happy accident, it is a dead gate — the slack is exactly how far each could regress before anyone was told.

All boards are now measured in one run, in one build configuration. The five boards re-baselined by hand in #798 reproduce to **-0.1%**, which is both what the metric is supposed to do and evidence the measurement was fine all along — it was the reference numbers that were wrong.

### Coverage: 11 chips → 25

The fixture had one checked-in `memory.x` at flash `0x08000000`, so the gate could only ever measure STM32 parts. `build.rs` now takes the origins from the environment and the gate links one fixture per memory map.

| Fixture | ISA | Chips brought in |
| --- | --- | --- |
| `stm32` | Cortex-M | *(also `stm32f405`, `stm32f411ceu6`, `stm32f767` — see below)* |
| `nrf` | Cortex-M | `nrf52832`, `nrf52840`, `nrf5340`, `nrf54l15` |
| `kinetis` | Cortex-M | `mkw41z4` |
| `rp2xxx` | Cortex-M | `rp2040`, `rp2350` |
| `esp32c3` | RISC-V | `esp32c3` |
| `esp32` / `esp32s3` | Xtensa | `esp32`, `esp32s3`, `esp32s3-zero` |

Worse than the documented hole: the hand-kept `BOARDS` list had silently stopped covering `stm32f405`, `stm32f411ceu6`, `stm32f767` and `rp2350`, and **none of them appeared in the "not covered" note either** — the gate read as complete while measuring nothing for them. Coverage is now derived from `configs/chips/`, and a chip no fixture matches is a hard error. `WAIVED` is empty for the first time.

Fixture matching is keyed on `(arch, flash, ram)`, not the memory map alone: the C3 and the S3 both boot at `0x42000000`, and keying on addresses would have handed the S3 a RISC-V ELF. There is a test for it.

### Acting on a regression instead of logging it

- The issue is a **live status page** — each run rewrites the body with the current report, and a comment is posted only when the set of failing boards actually changes. That is the spam fix.
- **A green run closes it.** Nothing closed these before, which is why an old regression stayed open looking like the new one.
- Issues are **assigned on creation**.
- A baseline that is too high now **fails**, so wins get locked in instead of turning into slack.
- A gate that dies before measuring is titled as a broken gate, not filed as a throughput regression against a stale signature.
- `--require-all` (what CI passes) means a chip that could not be measured is never a green run.

## Related Issues

Fixes #778 — and the workflow now closes it automatically on the first green run.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] New peripheral/architecture support — RISC-V and Xtensa perf fixtures
- [ ] Documentation update

## Checklist

- [x] I have run `cargo fmt` and `cargo clippy` (both fixture crates, on their cross targets).
- [x] I have added tests that prove my fix is effective or that my feature works — `scripts/perf/test_board_perf.py`, 10 tests covering the property that actually failed: a chip cannot leave the gate without someone being told. They run in `pr-gate` alongside `scripts/ci/test_board_matrix.py`, which had nine passing tests that no workflow had ever executed.
- [x] I have updated the documentation accordingly — the module docstring and the workflow header carry the rationale.
- [x] My changes generate no new warnings.

### What was verified locally

- Full matrix measured against the committed baselines twice: once before merging `main`, and again after. **22 of 22** locally-measurable boards reproduce at **±0.0%** both times.
- The second run matters because `main` moved 8 commits during review, touching `crates/core/src/cpu/riscv.rs`, `cpu/xtensa_lx7.rs` and `core/src/lib.rs` — the step path these baselines measure. None of it moved per-step cost on any board, so the committed numbers still hold. (`nrf5340` showed +0.6% on one intermediate run and returned exactly to baseline on the next; that was run-to-run noise, not drift.)
- The five boards hand-baselined in #798 reproduce to **-0.1%** against their *pre-existing* values, which is the evidence that the metric was fine and the reference numbers were not.
- An injected 45% regression fails with the correct status JSON.
- ESP32-C3 verified by **tracing execution**, not by trusting the number: a tight 3-instruction ALU loop at `0x42000134-138`, never reaching `DefaultExceptionHandler` or `abort`. Its 393.6 Ir/step is genuinely below the Cortex-M boards because the RISC-V step path is cheaper, not because the fixture traps.
- The skip path exercised with the `esp` toolchain genuinely absent: the three Xtensa chips are reported as NOT measured, and `--require-all` fails the run rather than passing it green.

### What CI runs first

The **Xtensa leg** is the one part not exercised here: `espup install` needs GitHub release downloads, which this sandbox's egress policy blocks (403). The simulator side is proven — it runs the committed Xtensa tier-1 ELF fine — so what CI exercises first is the fixture crate compiling. That is why the whole Xtensa path is built to degrade to a *reported* skip rather than a broken gate: `continue-on-error` on the espup step, `optional` on the fixture, and a loud line in the report either way.


w1ne:
`pr-gate` is red on the **silicon-drift gate**, and it fails identically on `main`. Not this PR.

I ran `python3 scripts/generate_validation_status.py --check --drift` against a pristine `origin/main` worktree (`e0ef4f6`): exit 1, same five boards — `stm32h563`, `nucleo-l476rg`, `nucleo-l073rz`, `stm32f103`, `stm32f407`.

It is a date rollover. `validation/manifest.yaml` carries a `drift_ack` dated **2026-08-06**; today is the 7th, so the model dates re-generate to `2026-08-07` and the ack no longer covers them. Any PR built today hits this.

Resolving it needs either a fresh silicon re-capture or a new dated `drift_ack` — an assertion that the models still match real hardware despite having changed. That is a hardware-truth call, not something to stamp in passing to turn a required check green, and it belongs in its own change rather than inside a perf PR. Flagging rather than fixing.

The rest of the PR's checks are green, and the perf step added here passes.

---

Separately, worth recording since #830 landed mid-review: **the ARM numbers in this PR do not measure the path #830 fixed.**

#830 removed three clamps pinning the CPU quantum to one instruction (9–16× native, 4.8 → 75.3 M steps/s on L476). Re-measured against it, all 22 boards moved **+0.2%…+0.4%** — baselines updated in `8cdf824`.

The reason a 16× win reads as +0.3%: `labwired run` drives ARM through `run_firmware_arm`, a `for i in 0..limit { machine.step() }` loop. `run.rs` itself notes that this call "pins the batch to one instruction", so #830's win can only appear through `Machine::run` / `step_batch` — which this gate never enters on ARM.

That narrows the gate's stated rationale. It claims the number transfers because the browser runs the same engine; but the browser's entry is `Sim::step_batch` → `machine.advance(AdvanceRequest::run)`, the batched path. After #830 those are no longer the same path on ARM.

The gate still catches what it was built for — a per-instruction cost on a shared path, the `std::env::var`-in-`CortexM::step` class — and every committed baseline is a true measurement of that path. What it cannot see is a regression in the batch orchestration itself, e.g. a fourth clamp of the kind #830 deleted, which on ARM is now the code the browser actually runs. Covering it needs a batched ARM driver (the RISC-V side already has one — `run_firmware_riscv_batched`, which is why `esp32c3` came in at -0.0%). Left as a follow-up so it is a known gap rather than a silent one.

---

PR #829, opened 2026-08-07
fix(nrf52): a GPIO edge must not erase a RADIO packet's air time

w1ne:
## What broke

Two fidelity gates went red on `main`:

- `radio_ble_1mbit_bit_time_scales_with_length` — `Ble_1Mbit L=1: END at 2, model expected 34`
- `radio_tx_end_walk1_vs_sched512_cycle_identity` — `delta=31`

`EVENTS_END` fired at cycle 2 for every packet length. The nRF52840 twin was not
simulating BLE 1 Mbit transmission time at all — it was raising END on the
EasyDMA cycle and calling it done. The test's model is right: at 1 Mbit one byte
is 8 µs, air bytes are `LENGTH + 3` CRC, and the harness clocks 1 cycle = 1 µs,
so `END = 2 + (LENGTH + 3) * 8` is the correct expectation. The model was wrong,
not the gate.

First bad commit: `4a37fda3` (#823), whose entire diff is two YAML files.

Run the lane with the feature or it lies:

```
cargo test -p labwired-core --features event-scheduler --test nrf52_timer_walk_differential
```

A plain `cargo test --test …` on that file runs ZERO tests and exits 0.

## Mechanism

#823 added a `board_io` block to `configs/systems/nrf52840-dk.yaml`. Bisecting
inside the YAML: the LEDs are innocent, a single **button** entry reproduces it.

1. `attach_board_io_buttons` settles each contact's released level through
   `drive_input_bit` at attach, so gpio0 P0.11 reads 1 before cycle 0.
   `last_gpio_in` was seeded `[0; 2]`, so the first per-tick GPIO edge pass
   diffed 1 against 0 and reported **a rising edge that nothing caused**.

2. On any edge the bus harvested `take_scheduled_events()` from **every**
   `uses_scheduler()` peripheral. Only GPIOTE can latch anything from a GPIO
   edge — it is the sole `observe_gpio_change` implementor — so for everyone
   else that harvest is over-reach, and it is not free: `take_scheduled_events`
   is a *query* of live state, not a one-shot take. The RADIO, mid-TXEN chain
   with a wake already in the heap, handed out a second one. The deadlines
   differ (`current_cycle + 1 + delay`), so `EventScheduler::schedule`'s
   byte-identical `(idx, token, deadline)` dedup could not collapse it.

Traced, at interval 1 (`due@2` carries the duplicate):

```
[sched] collect idx=21 name=radio deadline=1 tok=2     <- TXEN
[sched] collect idx=21 name=radio deadline=1 tok=2     <- TASKS_START
[sched] collect idx=21 name=radio deadline=2 tok=2     <- the phantom-edge harvest
[sched] due@2: [(21, 2), (21, 2)]
[radio] tick_with_bus -> TXDMA len=1 mode=3 rem=Some(32)
[sched] reschedule idx=21 at=33 tok=2                  <- correct END
[radio] on_event rem=Some(1) -> tick() -> END          <- the duplicate, same drain
  cyc=2 END=Ok(1)
```

The duplicate re-entered `on_event` while the air-time countdown was pinned at
1 and raised ADDRESS/PAYLOAD/END on the spot. Only the interval-1 lane runs the
phase-1 tick that owns the edge pass, which is why the two lanes disagreed by
exactly 31 cycles.

## Fix

Two defects, each gated by its own new test.

- **`observe_gpio_change` now returns whether it LATCHED work**, and the bus
  harvests only from those. GPIOTE reports it; the default is `false` — a model
  that does not observe GPIO cannot have latched anything.
- **`last_gpio_in` is `Option<[u32; 2]>`.** The first edge pass ADOPTS the live
  levels and reports nothing. An edge is a transition; a level the outside world
  already holds was never a press. `Option` rather than a companion flag so no
  construction site can spell "not yet sampled" as "sampled zero".

Either fix alone turns the two original tests green — the second is **not**
redundant, and I checked rather than assumed:

| | `radio_air_time_survives_a_gpio_edge_mid_transmission` | `board_io_button_boot_level_is_not_a_gpio_edge` |
|---|---|---|
| harvest fix only | pass | **FAIL** |
| baseline fix only | **FAIL** | pass |
| both | pass | pass |

A real button press mid-TX still corrupted air time with only the baseline fix.
The boot level still fired a GPIOTE `EVENTS_IN` nothing caused with only the
harvest fix.

## Red → green

Both new tests were watched failing first, on the current model:

```
board_io_button_boot_level_is_not_a_gpio_edge ... FAILED   (EVENTS_IN[0] == 1)
radio_air_time_survives_a_gpio_edge_mid_transmission ... FAILED
    a GPIO edge mid-TX must not shorten Ble_1Mbit air time: END at 2, expected 90
```

After:

```
running 8 tests
test board_io_button_boot_level_is_not_a_gpio_edge ... ok
test radio_air_time_survives_a_gpio_edge_mid_transmission ... ok
test radio_ble_1mbit_bit_time_scales_with_length ... ok
test radio_tx_end_walk1_vs_sched512_cycle_identity ... ok
test rtc0_compare_walk1_vs_sched512_cycle_identity ... ok
test rtc_counter_poll_advances_under_sched_tick512 ... ok
test rtc_counter_poll_walk1_vs_sched512_identity ... ok
test timer0_compare_walk1_vs_sched512_cycle_identity ... ok
test result: ok. 8 passed; 0 failed
```

`END` at L=1: 2 → 33. At L=8: 2 → 90. Walk-vs-scheduler delta: 31 → 0. No
assertion was weakened; the two originals are untouched.

Regression sweep, local:

- `cargo test --workspace --lib` — 3110 passed, 0 failed
- the exact new `core-integrity` feature-gate command — 2911 passed, 0 failed
- the exact new `pr-scheduler-observable` command — 27 passed, 0 failed
- `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --all` applied

## The gate gap — why this shipped green

`crates/core/tests/` holds 23 integration files behind
`#![cfg(feature = "event-scheduler")]`. With the feature off they do not skip —
they report **GREEN with zero tests run**. Eighteen of the 23, including
`nrf52_timer_walk_differential`, ran **only** in `core-full`
(schedule / `workflow_dispatch` / `[full-ci]`). Nothing on a PR or on push to
main was checking them, so #823 merged with every required check green.

Proportionate, given 2 vCPU runners that are already saturated:

- **The cheap half runs pre-merge.** Eight sub-2s lanes (nRF52 ×3, nRF54L15,
  RP2040 ×2, F103, SysTick) added to `pr-scheduler-observable`, and to
  `core-integrity` as the push backstop. Both jobs already build this crate with
  this feature, so the marginal cost is ~3s of test runtime plus the links —
  measured, not estimated.
- **The expensive half stays nightly, but says so.** New static ratchet
  `tests::scheduler_lane_coverage`: a scheduler-gated integration test must be
  named in a merge-gating lane of `core-ci.yml`, or listed in `NIGHTLY_ONLY`
  with a written reason. Coverage goes 5/23 → 13/23; the ten excused entries are
  the four 9–19s STM32 walk lanes, the 43.6s nRF54L15 idle-fast-forward measurement (the wall clock *is*
  the assertion), and five files that are entirely `#[ignore]`d — a lane running
  zero of them would be vacuously green, which is worse than nightly.

The ratchet is a file scan on the existing `--lib` lane: no new job, no runner
time. It was watched failing too — deleting `--test nrf52_timer_walk_differential`
from the YAML turns it red naming exactly that test. Two companion assertions
keep the scan itself honest (`full:` must stay the last job; a `NIGHTLY_ONLY`
entry must name a real gated file that no lane already covers).

I did not make `core-full` run on push. It is a 45-minute job and the useful
half of it is now covered by lanes that cost seconds.


w1ne:
Superseded by #863 — closing.

Verified on `main` rather than assumed; all three pieces are there:

- `radio_air_time_survives_a_gpio_edge_mid_transmission` is present in `nrf52_timer_walk_differential.rs`.
- The `assert_eq!` my earlier conflict resolution truncated is restored (`itself as a GPIO edge" );` closing cleanly).
- `#[cfg(test)]` sits on `pub mod simctl_machine;` again — the concatenated merge had displaced it onto the new module, which pushed `simctl_machine` into the lib target and produced 10 dead-code errors under `-D warnings`.

The GPIOTE drift rationale is in `validation/manifest.yaml` alongside main's RADIO air-time and TIMER re-arm notes — both sets kept, digests re-stamped.

Branch head is not an ancestor of main only because fixes were pushed after the batch was cut.

PR #831, opened 2026-08-07
fix(console): tap the console the board's USB socket is actually wired to

w1ne:
## The coupling

An ESP32-C3/S3 has **two** consoles: UART0 and the chip's own USB-Serial-JTAG. Which one carries `Serial` to the developer's cable is a **board** fact — the board contract's `deploy.usb`:

| board's USB socket | console | `ARDUINO_USB_CDC_ON_BOOT` |
|---|---|---|
| `native` (esp32-c3-supermini, esp32-s3-zero) | USB-Serial-JTAG | 1 |
| bridge IC (esp32, adafruit-feather-esp32-v2) | UART0 | 0 |

w1ne/labwired#1427 makes the **build** derive from that field. This makes the **twin** derive from the same field, so the two halves of one board fact cannot fork.

## What was wrong

The wasm entry decided the console four separate times, each with the same hand-written string compare and the same silent fallback:

```rust
if !bus.attach_uart_tx_sink_named(name, sink, false) {
    bus.attach_uart_tx_sink(sink, false);      // ... any UART at all
}
```

`debug_uart: usb_serial_jtag` was honoured on the C3 ROM path and **quietly answered with UART0 everywhere else**. For a twin that is the worst available answer: the pane fills with plausible text while claiming `Serial` is on pins the board never uses.

## What this does

- **ONE home.** `labwired_core::console` parses the declaration (`HostConsole::from_manifest`) and pairs it with the console that therefore has no connector (`ConsoleCapture::for_manifest`). Every construction path asks there.
- **Refuses, never substitutes.** `SystemBus::attach_host_console` errors when the bus cannot provide the declared console. A declarative register stub at `0x6004_3000` does not count either — it never drains a byte.
- **One tap, not two.** Merging both consoles would make every run "work", which is exactly wrong: no real board hands you two consoles on one cable, so a merged pane would *hide* the CDC-on-boot bug instead of reproducing it. It also breaks the C3 faithful ROM path, where the mask ROM prints its banner to **both** consoles — a merged buffer renders every ROM character twice.
- **But not silent.** The untapped console is still recorded. `console_mismatch()` / `unheard_console_output()` report what the firmware said there, minus the prefix both consoles received (so the ROM banner never raises a false alarm). A real board shows an empty pane; the twin shows the same empty pane **and the reason**.
- **Nothing that ships today changes.** The paths that run no mask ROM (S3 fast-boot, C3 bare-ELF) still capture both consoles when the manifest declares nothing — there is no banner to double there, and no bug to fix.

## Evidence, both directions

New fixture `crates/wasm/tests/fixtures/esp32c3-usb-cdc-console-flash.bin`: a real hosted-PlatformIO arduino build for `esp32-c3-supermini` that drives **both** consoles from one image (`Serial.println("LW_UART0_TICK")` + `HWCDC UsbCdc; UsbCdc.println(...)`), booted through the genuine mask ROM.

| manifest | pane shows | pane must NOT show | unheard |
|---|---|---|---|
| `debug_uart: usb_serial_jtag` | ROM banner over USB-Serial-JTAG | `LW_UART0_TICK` | `LW_UART0_TICK` (non-vacuous: UART0 demonstrably printed) |
| *(undeclared — every lab today)* | `LW_UART0_TICK`, ROM banner **exactly once** | — | empty |

```
cargo test -p labwired-wasm --release console_tap_tests -- --ignored
  native_usb_board_shows_the_usb_console ............ ok
  bridge_console_board_still_shows_uart0 ............ ok
  output_on_the_untapped_console_is_reported_not_lost ok

cargo test -p labwired-core --test host_console_selection ... 6 passed
cargo test -p labwired-core --features event-scheduler ...... no new failures
```

Watched red first: `crates/core/tests/host_console_selection.rs` was run against the restored old fallback — `a_chip_without_a_usb_serial_jtag_block_refuses_that_console` and `a_uart_this_bus_does_not_have_is_refused` both FAILED (the sink came back attached to a UART), and pass only with the refusal in place. `unheard_console_output` did not exist at all, so the ROM-boot tests failed to compile before the fix.

Baseline: `nrf52_timer_walk_differential::radio_ble_1mbit_bit_time_scales_with_length` and `radio_tx_end_walk1_vs_sched512_cycle_identity` are RED on pristine `origin/main` (verified by stashing this branch). Not from this change.

## ⚠️ This is necessary and NOT sufficient — do not deploy #1427 on it

Closing the tap does **not** yet make a CDC-on-boot Arduino build print in the twin. `arduino-esp32`'s HWCDC — the driver a CDC-on-boot build binds to `Serial` — is entirely interrupt-driven:

- `HWCDC::write` only enqueues when `isCDC_Connected()`, which calls `usb_serial_jtag_is_connected()` — a **SOF-frame watchdog on `INT_RAW.SOF`**.
- The ring buffer only reaches the TX FIFO from the **`SERIAL_IN_EMPTY` ISR** (`hw_cdc_isr_handler`, which is also the only thing that sets `connected = true`).

Our `UsbSerialJtag` model is a polling-only byte sink: `EP1_CONF` permanently `WR_DONE | SERIAL_IN_EP_DATA_FREE`, **no interrupt registers at all**. That is why the mask ROM's `usb_uart_tx_one_char` busy-poll works through it, and why this fixture's `LW_USBCDC_TICK` never appears — measured, not assumed: zero bytes at the USB model over 400M steps while the same image's UART0 output arrives normally.

**Follow-up, and a hard prerequisite for deploying #1427: model the attached USB host — `INT_RAW`/`INT_ENA`/`INT_CLR`, a SOF every 1 ms, `SERIAL_IN_EMPTY` on host pickup, and matrix IRQ delivery (opt-in per chip; the C3 and S3 source numbers differ).** The IRQ plumbing is already data-driven (`matrix_irq_sources_into`, cf. `Esp32c3Rmt`).

## Companion change (app repo, ships together)

Core cannot read the board contract. The seam is the run manifest, and the app must derive it from the same field as the build flags — in `usbSerialBuildFlags`'s file, not a second place:

```ts
// boards.ts, next to usbSerialBuildFlags — ONE derivation of deploy.usb
export function hostConsole(deploy: BoardDeploy | undefined): string | null {
  if (!deploy) return null;
  if (deploy.espChip !== 'esp32c3' && deploy.espChip !== 'esp32s3') return null;
  return deploy.usb === 'native' ? 'usb_serial_jtag' : 'uart0';
}
```

emitted by `buildSystemYaml` (compile/emitters.ts) as a top-level `debug_uart:` line. Note `carryEngineFlagLines` already skips a key the emitter produced, so bundled-lab carry-over stays correct. **Check the S3 case before enabling it there** — S3 fast-boot currently captures both consoles, so declaring one narrows it.

## Deploy

Requires a **wasm rebuild** to reach the browser. Not attempted here.


w1ne:
CI note: `pr-gate` fails on the silicon drift-watch —

```
  stm32h563 nucleo-l476rg nucleo-l073rz stm32f103 stm32f407
```

That is the `drift_ack: 2026-08-06` in `validation/manifest.yaml` expiring at the UTC date rollover, not anything in this branch: `main` itself is red right now for the same reason (`Rust Core CI` on `e0ef4f6f`, job `core-integrity`, same ERROR plus `docs/boards/VALIDATION_STATUS.md is out of date`). Any core change touching `crates/core` today trips it. Deliberately not re-acked here — that is a silicon attestation, not a CI knob.

Locally, on this branch: `cargo test -p labwired-core --features event-scheduler` has no new failures; `nrf52_timer_walk_differential::{radio_ble_1mbit_bit_time_scales_with_length, radio_tx_end_walk1_vs_sched512_cycle_identity}` are red on pristine `origin/main` too (verified by stashing this branch and re-running).

PR #832, opened 2026-08-07
feat(usb_serial_jtag): model the interrupt surface so CDC-on-boot builds print

w1ne:
## The bug

An ESP32-C3 SuperMini / ESP32-S3-Zero wires its USB-C socket to the chip's own
USB-Serial-JTAG block, not to UART0. Firmware for those boards is built with
`-DARDUINO_USB_CDC_ON_BOOT=1 -DARDUINO_USB_MODE=1`, which makes Arduino's
`Serial` an interrupt-driven `HWCDC`. In such an image `HardwareSerial` is not
linked at all — there is no UART0 console to fall back to.

Our model had no interrupt surface whatsoever. `INT_RAW` / `INT_ST` / `INT_ENA` /
`INT_CLR` all read literal 0, `INT_ENA` dropped writes, and nothing was ever
asserted into the interrupt matrix. A CDC-on-boot build emitted **zero bytes**.

That is the trust inversion the twin exists to prevent: app PR
w1ne/labwired#1427 makes the build do the right thing for the real board, and
would have simultaneously blinded the simulator.

### Root cause: SOF, not SERIAL_IN_EMPTY

`HWCDC` maintains `s_usb_serial_jtag_conn_status` **entirely** from
`USB_SERIAL_JTAG.int_raw`'s SOF bit, polled by a 1 kHz FreeRTOS tick hook
(`HWCDC.cpp:47-78`). The hook reads `INT_RAW` directly, so `SOF_INT_ENA` is
irrelevant — the raw bit must be set regardless of enable. It starts `true` with
a 5-tick budget (`pdMS_TO_TICKS(5)`); with no SOF the flag latches false ~6 ms
after system init, long before `setup()` runs.

Once false, `HWCDC::write` (`HWCDC.cpp:419-421`) routes every byte to
`flushTXBuffer()`, which cycles it through the ring buffer and **discards** it —
while still returning `size`, so the sketch sees success. **A twin that never
raises SOF is modelling an UNPLUGGED board**, and no amount of correct
`SERIAL_IN_EMPTY` logic produces output.

## What changed

`crates/core/src/peripherals/esp32s3/usb_serial_jtag.rs`:

* **SOF** at the 1 kHz USB full-speed frame rate, derived lazily from the shared
  `CycleClock`. The driver only ever reads the bit and `read` takes `&self`, so
  this is the `Cell`-held lazy pattern the `Peripheral` trait documents. Costs
  nothing per cycle — no walk tick, no scheduled wake.
* **`INT_RAW`** = latched (W1C) bits OR live level conditions.
  **`INT_ST`** derived as `INT_RAW & INT_ENA`, never separate storage.
  **`INT_ENA`** real read/write storage (the LL does read-modify-write).
  **`INT_CLR`** write-1-to-clear; writing 0 to a bit does not clear it.
* **`SERIAL_IN_EMPTY` as a LEVEL that resets to 1.** The register header types it
  `R/WTC/SS, default: 1`. That reset value is the *ignition* — `begin()` writes
  `INT_ENA=0x20C` and `INT_ST` bit 3 goes high with no USB traffic at all. The
  re-assertion after W1C is the *pump* for every byte after the first. It does
  not spin, because the driver disables the enable when its ring buffer runs dry.
* **Measured back-pressure.** `WR_DONE` drops `DATA_FREE` and `SERIAL_IN_EMPTY`
  together; host pickup (~235 us) returns both together; a store to EP1 while
  `DATA_FREE == 0` is ignored. Reporting "always writable" would have been easier
  and dishonest — `usb_serial_jtag_ll_write_txfifo` breaks its loop on that bit.
* **`CONF0`** as storage (firmware read-modify-writes the PHY bits).
* **Assertion into the interrupt matrix** via `matrix_irq_sources_into`.

### Per-chip, opt-in, never defaulted

|                | ESP32-C3     | ESP32-S3     |
|----------------|--------------|--------------|
| Base           | `0x60043000` | `0x60038000` |
| Matrix source  | 26           | 96           |

These **differ**, which is exactly the shared-model trap. `UsbSerialJtag::new()`
wires **no** interrupt source; callers name their chip via `new_esp32c3()` /
`new_esp32s3()`. An S3 assumption cannot reach C3 silicon by omission.

Source ids are `ETS_USB_SERIAL_JTAG_INTR_SOURCE` counted in each chip's own
`periph_defs.h` (unnumbered enumerators). Cross-checked independently: C3 = 26 is
visible in the compiled image (`HWCDC::begin` emits `li a0,26` right before
`jal esp_intr_alloc`, near `lui a5,0x60043`); the S3 enum walk that yields 96 also
reproduces `ETS_I2C_EXT0=42`, `ETS_I2C_EXT1=43`, `ETS_LEDC=35`, `ETS_LCD_CAM=24`,
which this crate already pins elsewhere.

Register offsets and bit positions are **identical** on both chips — verified by
diffing the two `usb_serial_jtag_reg.h` copies, not assumed. Base addresses are
*not* constants in the model (the chip YAML owns them; see
`tests::yaml_owned_base_contract`).

### Failing first

`crates/core/tests/esp32c3_usb_cdc_console.rs` boots a **real** PlatformIO
Arduino image for `board = esp32-c3-supermini`, `platform = espressif32@7.0.1`,
built with `-DARDUINO_USB_CDC_ON_BOOT=1 -DARDUINO_USB_MODE=1`. On unmodified
`origin/main`:

```
test cdc_on_boot_build_emits_loop_output_through_usb_serial_jtag ... FAILED
no CDC loop output after 120000000 steps.
USB_SERIAL_JTAG sink (0 bytes): ""
UART0 sink (0 bytes, tail): ""
```

**The control is what makes that meaningful.** The identical sketch built with
`-DARDUINO_USB_CDC_ON_BOOT=0` boots through the same harness and prints normally
(`LW_CDC_SETUP`, `LW_CDC_LOOP 0..14`), so the ROM boot, bootloader, Arduino
startup and `loop()` all work. The two images differ in the console and nothing
else: the CDC image links `HWCDC` and **zero** `HardwareSerial` symbols; the
control links `HardwareSerial` and **zero** `HWCDC`.

The assertion is on the **second** loop iteration, not the `setup()` line — a
sketch whose `loop()` is dead can still print from `setup()`.

### Passing after

```
test cdc_on_boot_build_emits_loop_output_through_usb_serial_jtag ...
    CDC loop output observed after 10200000 steps (44 bytes captured)      ok
test cdc_sink_bytes_all_came_through_the_ep1_fifo ...                      ok
test output_stops_when_the_host_stops_sending_sof ...
    SOF withdrawn: console went silent at 44 bytes                          ok
test uart0_build_puts_no_application_output_on_the_cdc_sink ...
    control: 0 bytes at the CDC sink, none of them application output       ok
test uart0_control_build_emits_loop_output ...                             ok

test result: ok. 5 passed; 0 failed
```

**10.2M steps — comfortably under the 20M default `max_steps`.**

### Negative controls

1. **`uart0_build_puts_no_application_output_on_the_cdc_sink`** — the UART0 build
   puts no application output on the CDC sink (0 bytes). Output there would mean
   `Serial` is being short-circuited into the sink rather than modelled.
2. **`output_stops_when_the_host_stops_sending_sof`** — with the SOF generator
   switched off, the console goes silent and stays silent for a further 40M steps
   (the sketch prints every 50 ms, so it would have produced many more lines).
   Bytes continuing to flow would mean `isPlugged()` is not actually consulted
   and the FIFO path is being bypassed.
3. **Byte conservation** — bytes accepted at `EP1.RDWR_BYTE` == bytes emitted ==
   bytes in the sink, asserted both as a unit test over 2560 bytes and end to end.

## Per-chip coverage — what I did NOT verify

**ESP32-C3: verified end to end.** Real Arduino CDC-on-boot image, real mask ROM,
output through the sink.

**ESP32-S3: NOT verified end to end.** No S3 firmware was booted against this
model and no S3 hardware was attached. Every S3 claim is header-derived (base,
source id, identical register layout) and mechanically cross-checked, but
interrupt delivery for source 96 through the S3 Xtensa intmatrix has never been
exercised by a running S3 image. The repo has no Arduino flash-boot harness for
S3 equivalent to the C3 one, so building that gate is follow-up work. This is
stated in the module docs too, so the next reader does not mistake C3 evidence
for S3 evidence.

**RX is scoped out, explicitly.** `SERIAL_OUT_RECV_PKT`, the RX FIFO, the `EP1`
read path and `SERIAL_OUT_EP_DATA_AVAIL` are all modelled and unit-tested via
`inject_rx`, but nothing on the bus feeds them yet — injected serial still
reaches UART0 only. Wiring the console's input side to the CDC endpoint is left
out on purpose rather than being quietly absent.

## Merge-order note

This branch is cut from `origin/main` and does **not** stack on core#831. The
end-to-end test attaches the CDC sink itself through `RomBootOpts::usb_serial_sink`,
so it stands alone. #831 is what makes a *user's* board tap the right console; the
two are independent and can merge in either order.

## Known blocker, untouched

`validation/manifest.yaml` has `drift_ack: 2026-08-06`, which expired at UTC
rollover — `main` is red for that reason independently of this PR. That is a
silicon attestation only the repo owner can make, so I have deliberately not
touched or renewed it.


w1ne:
## CI status and local gate evidence

`pr-gate` is red, but **not on anything this PR touches**. Its first step
`Validation status — regen check + drift gate` fails and short-circuits the
rest, so clippy and the test steps were **skipped**, not failed:

```
failure  Validation status — regen check + drift gate
skipped  Generated artifacts — regenerate and assert no drift
skipped  Clippy (default-members)
skipped  Unit tests (default-members lib)
skipped  Clippy + build the browser layer (crates/wasm)
skipped  Unit tests (browser layer)
skipped  Walk-deletion derivation (PR-fast structural arm)
```

That step fails because `validation/manifest.yaml` carries
`drift_ack: 2026-08-06`, which expired at UTC rollover. It lists six boards —
`stm32h563`, `nucleo-l476rg`, `nucleo-l073rz`, `stm32f103`, `stm32f407`,
`esp32s3` — five of which this PR does not touch at all. `main` fails the
identical step in its own `core-integrity` job (run 31147921018), so this is a
pre-existing blocker. `drift_ack` is a silicon attestation only the repo owner
can make, so it has deliberately been left alone here.

Since CI skipped them, the CI commands were run locally instead:

| command | result |
|---|---|
| `cargo clippy --all-targets -- -D warnings` | clean |
| `cargo clippy -p labwired-wasm --all-targets -- -D warnings` | clean |
| `cargo test --lib` (default-members, default features) | **2848 passed, 0 failed** |
| `cargo test -p labwired-core --features event-scheduler --lib` | **2837 passed, 0 failed** |
| `cargo fmt --all -- --check` | clean |

The other PR checks that did run are green, including
`pr-scheduler-observable` (the walk-deletion observable arm, which is the gate
most exposed to this PR's `needs_legacy_walk()` change) and `esp32c3-blinky`.

### Baseline note

The two tests red on `main`'s `core-full` today are
`radio_ble_1mbit_bit_time_scales_with_length` and
`radio_tx_end_walk1_vs_sched512_cycle_identity`, both in
`crates/core/tests/nrf52_timer_walk_differential.rs` — nRF52 radio timing,
unrelated to USB_SERIAL_JTAG.

w1ne:
## Baseline vs branch: zero new failures

Full local sweep, `cargo test -p labwired-core --features event-scheduler --no-fail-fast`:

```
test binaries completed: 134
failed suites:             1
individual tests:       3275 passed, 2 failed
```

The two failures are:

```
crates/core/tests/nrf52_timer_walk_differential.rs
    radio_ble_1mbit_bit_time_scales_with_length
    radio_tx_end_walk1_vs_sched512_cycle_identity
```

These are **exactly** the two tests already red on `main` (its own `core-full`
job, run 31147921018) — nRF52 radio timing, nothing to do with
USB_SERIAL_JTAG. **No new failure is attributable to this PR.**

The sweep stopped at binary 134 of 158 because the machine ran out of disk,
not because of a failure; the remaining binaries are alphabetically after
`stm32*`/`systick*`. Everything the PR plausibly touches — the C3 boot path,
the S3 peripheral set, the walk-deletion and scheduler contracts, the display
and e2e lanes — ran and is green.

One suite failure earlier in the run was an environment artifact, not a
regression, and is worth recording: `e2e_epaper_tricolor` shells out to build
its bare-metal ELF, then looks for it at a hardcoded
`crates/core/../../target/<triple>/release/...`. Under a custom
`CARGO_TARGET_DIR` the child build writes elsewhere, so the test fails with
"ELF not found ... after build" even though the build succeeded. With the path
resolved it passes in 154 s. Unrelated to this PR, but it means that test
cannot pass under a non-default `CARGO_TARGET_DIR`.

w1ne:
Superseded by #863 — closing.

This PR's work is on `main`, landed through the core integration batch:

- The USB_SERIAL_JTAG interrupt surface (INT_RAW/ST/ENA/CLR + matrix source id) is in the merged tree.
- The match-guard form is there too — `main` has `OFF_EP1 if lane == 0 && self.data_free() =>` and `OFF_EP1_CONF if lane == 0 && ... =>`, so the repo stays clean under clippy 1.95, which is what CI pins (1.97 no longer reports `collapsible_match`, which is why it looked green locally).
- `esp32c3_usb_cdc_console` now runs in the `core-integrity` release lane. #860's `scheduler_lane_coverage` gate caught that this file is `#![cfg(feature = "event-scheduler")]`, so without the feature it ran **zero tests and reported green** — this PR's headline behaviour had nothing checking it.

The branch head is not an ancestor of main only because fixes were pushed to it after the batch was cut, so GitHub could not auto-close it. Nothing here is unmerged.

PR #855, opened 2026-08-07
test(esp32c3): bind the BLE Pong fixture to the lab it claims to test

w1ne:
## The failure this closes

The published community lab **BLE Pong — two ESP32-C3s** (`c477f82961e86f601e7b908ae7e12311`) was broken in the browser — a dead panel — while three separate gates stayed green:

1. `world_esp32c3_ble_pong` — its docstring claimed it "deliberately uses the owner's published lab … because the whole point is that a real user's real sketch works". It did not. It read a **committed** flash image last rebuilt at #828, while the published sketch moved on repeatedly. It passed against firmware nobody ran.
2. `scripts/ci/lab-smoke.mjs` (superproject) — enumerates curated bundled labs addressed `?lab=<id>`. Community projects at `/p/<user>/<projectId>` are not covered at all.
3. Unit suites never boot a lab.

A test whose name and docstring promise coverage they do not have is worse than no test: it buys false confidence. This PR makes that specific gate track the real artifact, and adds the check that notices when the artifact moves.

## What refreshing the fixture found

Rebuilding the fixture from the then-current published sketch made the world test fail immediately:

```
nodeA tag=4 H ball=74,37 me=25 peer=24 score=0:0 rally=0
nodeB tag=5 G ball=64,32 me=25 peer=24 score=0:0 rally=0   <- spawn, never moved
panicked: nodeB ball never left spawn — no host frame ever arrived
```

Same engine, **old firmware green, new firmware red** — which isolates the firmware as the variable.

Mechanism, confirmed by soaking to ~3 s of simulated time: `PUBLISH_MS` was 700 ms against a 50 Hz game loop. The host's first advertisement is emitted at the end of `setup()` while `isHost` is still false, so it is a 5-byte guest-shaped frame carrying no world state — and the next one is 700 ms away. The guest had nothing to draw for the whole first period (`haveHostFrame` false → static fallback → dead panel). Given longer it repainted one stale snapshot ~15 frames running, drifting up to 70 px from the host.

The lab has since been fixed (dead reckoning: the host advertises velocity, the guest integrates between packets, with easing so corrections are spent over frames instead of jumping). The fixture here is that firmware, and the world test passes on it.

## Changes

- **`crates/core/tests/fixtures/esp32c3-ble-pong.ino`** (new) — the published sketch, verbatim.
- **`crates/core/tests/fixtures/esp32c3-ble-pong-flash.bin`** — rebuilt from it.
- **`crates/core/tests/fixture_ble_pong_provenance.rs`** (new) — pins the sha256 of both, plus a flash-layout guard. **Not `cfg`-gated**, so unlike the world test it runs on PRs.
- **`scripts/ci/check-published-lab-drift.sh`** (new) — fetches the live project and fails if it no longer matches the committed `.ino`, or if `source_code` and `firmware[0].source_code` have diverged (the lab's premise is one image on both chips). Fails closed: network failure and real drift both exit non-zero, with different messages.
- **`crates/core/tests/world_esp32c3_ble_pong.rs`** — docstring now states plainly that this is a frozen hash-bound fixture, not a live fetch; records what the cadence regression looked like; documents the budget blind spot; prints a serial tail on failure; adds a soak-only `PONG_SOAK_CYCLES` override (unset in CI — the constant is always the gate).
- **`.github/workflows/core-ci.yml`** — `pr-gate` gets the provenance test + drift check; `core-integrity` gets the drift check ahead of the release-gated world test.

## Why two mechanisms and not one

A pinned hash keeps the two committed files honest with each other but cannot see the third party — the owner editing the live project. A live fetch cannot run in CI's offline lanes. Neither is sufficient alone.

This is not theoretical: the drift check fired unprompted **during this work**, catching an edit to the live project made minutes earlier, and again caught `source_code` and `firmware[0].source_code` having diverged.

## Non-vacuity

Every assertion was made to fail first.

| Perturbation | Result |
|---|---|
| `.ino` one byte | `ble_pong_sketch_matches_its_recorded_hash` FAILED, others ok |
| `.bin` one byte | `ble_pong_flash_image_matches_its_recorded_hash` FAILED, others ok |
| `.bin` truncated to 0x8000 | layout guard FAILED — "too short to hold an app at 0x10000" |
| `.ino` one byte (drift script) | exit 1, "PUBLISHED LAB HAS DRIFTED" |
| nonexistent project id | exit 1, "FETCH FAILED (HTTP 404)" |
| `firmware[0]` diverged (local stub) | exit 1, "NO LONGER ONE IMAGE ON BOTH CHIPS" |
| `firmware[]` emptied (local stub) | exit 1, "second MCU lost its sketch" |
| **control**: unmodified stub | exit 0 |

The control earned its place: the first stub harness silently served nothing, and all three "proofs" passed for the wrong reason until the control caught it.

## Twin characterisation (recorded, no code change)

`PUBLISH_MS` 700 → 100 was a risk, because a fast advertising republish cadence is a known open twin defect (~20 ms provably stalls the simulated RW-BLE controller; 700 ms provably works). **100 ms was untested.** Measured here: it does not stall — see the run notes in the PR thread.


w1ne:
## Twin cadence characterisation — `PUBLISH_MS` 100 does **not** stall

`PUBLISH_MS` went 700 → 100 in the published lab. That was the open risk: a fast advertising republish cadence is a known twin defect (RWBLECNTL bit 25 / `r_lld_adv_stop`) where ~20 ms provably stalls the simulated RW-BLE controller and 700 ms provably works. **100 ms was untested.** Silicon is unaffected either way.

Measured on this branch, base `origin/main`, in an isolated worktree (no shared `target`):

| Run | Budget / node | Simulated | Republish periods @100 ms | Result |
|---|---|---|---|---|
| Gate (CI default) | 90M cycles | ~560 ms | ~1 | **2 passed** in 144 s |
| Soak | 480M cycles | ~3 s | ~30 | **2 passed** in 366 s |

Soak tail — the guest is tracking the host through a wall bounce and a scored point, not repainting a stale snapshot:

```
nodeA tag=4 H ball=6,3 v=-2,-1 me=25 peer=25 score=1:0 rally=0
nodeB tag=5 G ball=8,4 v=-2,-1 me=25 peer=25 score=1:0 rally=0

--- host ---            --- guest ---
H ball=20,10            G ball=22,11
H ball=18,9             G ball=20,10
H ball=16,8             G ball=19,10   <- 1 px step: the ease correcting
H ball=14,7             G ball=16,8
H ball=12,6             G ball=14,7
H ball=10,5             G ball=13,7    <- again
H ball=8,4              G ball=10,5
H ball=6,3              G ball=8,4
```

Host moves 2 px/frame; the guest dead-reckons at the same rate and closes residual error a pixel at a time. Final separation 2 px against an assertion threshold of 8. Score reached 1:0, so a full point — bounce, miss, respawn — completed inside the soak.

**Conclusion: no stall at 100 ms.** The known-bad region is somewhere in (20, 100]; this run does not narrow it further, and no bisect was needed because the landing value passes. Recording the numbers so the next person tightening the cadence knows 100 ms is measured, not assumed.

### Caveat worth keeping

The gate's 90M budget only covers ~140 ms of `loop()` time, so it sees roughly **one** republish. It would pass a firmware whose cadence had regressed to, say, 400 ms and only fail once the cadence exceeded the window entirely — which is exactly how the 700 ms regression presented ("no host frame ever arrived", indistinguishable from broken delivery). The soak override exists for that reason and CI never sets it. If cadence regressions become a recurring theme, the honest fix is a dedicated longer-budget lane, not a bigger constant on the merge gate.

PR #876, opened 2026-08-08
chore(census): count the three silent paths, measured when the machine runs

w1ne:
A **measurement**, not a fix. Nothing changes engine behaviour. It exists to size follow-up work and answer a pending decision about acceptable lab breakage.

Gated twice: Cargo feature `silent-census` (default off, not implied by any other feature) **and** `LABWIRED_CENSUS_OUT` naming an output path. Each gate verified alone: off-binary with the env var writes nothing; on-binary without it writes nothing.

> **This PR was wrong once and has been corrected in `8ac2fc13`.** The first version measured stubs at factory time and concluded `nvic` (22 labs) and `scb` (3) were accidentally stubbed. They are not: `system/cortex_m.rs` finds those entries and swaps in the real `Nvic`/`Scb`/`Dwt`, rewriting their names as it goes. A factory-time counter cannot see that. The correction and the reasoning are in the report so the next reader does not repeat it.

## Results

**(a) Dropped Cortex-M memory errors — cold.** 2 hits corpus-wide, both in `ci/dummy-memory-violation.yaml`, a fixture whose purpose is to provoke one. Cross-checked against the independent `fidelity::unmapped_mmio` log (10 bytes rejected, 8 of them *propagated* via the instruction-fetch `?`, not dropped).

This does **not** mean the gap is harmless — firmware that HardFaults on silicon still completes here with a green verdict. It means the *fix is cheap*: nothing in the corpus depends on the broken behaviour.

**(b2) Live stubs — 147, and every one was asked for.** Measured at `Machine::new`, holding a finished bus:

| count | asked for by |
|---|---|
| 120 | manifest `type: "stub"` |
| 18 | manifest `type: "nrf54l_*_stub"` (7 strings) |
| 7 | manifest `type: "icache"` — explicit `generic_factory` arm |
| 2 | manifest `type: "syscfg"` — explicit `generic_factory` arm |
| **0** | anything else |

**No accidental live stub exists.** (b1), the factory-fallthrough count, is retained separately and labelled as a different statistic — 163 across 47 labs.

The delta runs **both ways**, which is why the correction was worth making rather than patching: (b1) over-reported 25 entries replaced later, *and* under-reported 9 stubs installed by named factory arms that never reach `_other =>`. Clean labs went 19 → 25.

**(c) Undecoded register access — nearly cold.** 11 raw hits = 5 distinct register-level accesses, all writes: `fdcan@0x010`, `iwdg@0x008`, `twim@0x508/0x50c/0x510`. No clock-enable offset is undecoded on any running lab — the WB55 precedent does not reproduce. Declarative path measured zero.

## How a stub is detected

`Peripheral::as_any()` → `dyn Any::is::<StubPeripheral>()`. A `TypeId` comparison: no false positives on another model, no false negatives on a real stub. Name and base are *reported* so a human can find the entry, but are never the predicate — the replacement passes rewrite `name`, which is precisely how the first measurement failed.

The only production-type change is an `as_any` override on `StubPeripheral`, itself `#[cfg(feature = "silent-census")]`. Feature off, the type is byte-for-byte unchanged. Feature on, every `as_any` consumer in the tree immediately `downcast_ref`s to a concrete model a stub is not, so `Some(stub)` and `None` are indistinguishable.

## Other post-factory mutation passes, audited

`install_esp32c3_fast_boot` uses the same replace-mechanism and runs before `Machine::new`, so (b2) covers it. Xtensa builders and the C3 ROM injector are additive. `configure_riscv` mandates no peripherals — no RISC-V equivalent exists. The `.dev =` sites in `bus/tick.rs` and `esp32s3/gdma.rs` are transient mid-run swaps that restore. No construction pass runs after `Machine::new`; that limit is documented.

## Neutrality

**All 70 machines**, not a sample: `result.json` byte-identical between the off binary and an actively-recording on binary, zero exit-code mismatches, with all four counters firing somewhere in the compared set (58 pass / 10 fail / 2 config errors represented). Failing labs included deliberately — a counter that perturbed a failure would be as bad as one that perturbed a pass.

Gate built to fail first: `census_probe.rs::nvic_and_scb_trip_the_factory_counter_but_are_not_live_stubs` runs the real `from_config` on committed `configs/chips/stm32l073.yaml` and asserts both **are** stubs out of the factory, then **are not** after `configure_cortex_m`, with a non-vacuity check.

`cargo fmt --check` clean; clippy clean in both feature configurations; `labwired-core --lib` 2,799 (off) / 2,804 (on); `labwired-cli --lib` 113.

## Coverage, stated honestly

97 scripts discovered, **68 ran** (70 machines — two are multi-node), 27 skipped, each with a row and a reason: PlatformIO output, uncommitted C-Makefile ELFs, absent riscv32 target, three nested workspaces failing on this toolchain, two paths outside the repo.

## What this scopes

- Fault escalation is a small task, not a migration — nothing depends on the current behaviour.
- Stub enforcement needs an allowlist of four declared families, not an investigation.
- Undecoded-register enforcement starts from a five-entry allowlist.

## Incidental findings

`ci/benchmark.yaml` declares `max_steps` above the CLI cap; `stm32f103-integrated-test`'s manifest is missing `chip:`; `diff-smoke.yaml` hardcodes `/tmp/lw-deploy/...`; `scripts/example_smokes.sh`'s glob misses ~7 scripts plus all of `examples/ci/`. The `nvic`/`scb` entries in the chip YAMLs are inert now that the core installs both unconditionally — removing them is a separate task. 10 of 68 labs fail their own assertions pre-existing at this SHA, proven unrelated by the byte-identical failing-lab comparisons.


w1ne:
## Correction: finding (b) is overstated, and the `nvic`/`scb` conclusion in the description is wrong

I claimed 22 labs run with a stubbed NVIC and that NVIC masking is unmodelled on `stm32h563`/`h735`/`f767`. **That is incorrect.** A follow-up specifically to fix it found there was nothing to fix.

`system/cortex_m.rs:80-101` locates the entry matching `p.name == "nvic" || p.base == 0xE000_E100` and **replaces `p.dev` with the real `Nvic`**, then sets `bus.nvic`. Same for `Scb` at `0xE000_ED00` — and a real `peripherals/scb.rs` exists. Every ARM construction path calls it after `from_config`: the lab runner (`cli/commands/test.rs:1715`), the browser (`wasm/lib.rs:414`), plus `run.rs`, `machine.rs`, `debug_probe.rs`, `dap/adapter.rs`, `python/lib.rs`, `system/node.rs`.

Verified independently: a guard asserting `NVIC_ICER` masking works, that an enabled IRQ still fires, and that all 13 `nvic`-declaring chips get the real model — **passes on unmodified `main`**. It is not a vacuous test: the identical assertions run against a `from_config`-only bus fail on all 13 chips (`bus.nvic is None`, `ISER0_readback=0x00000000`, `SCB_CPUID=0x00000000`).

### The methodology flaw this exposes

**The census counts peripheral construction at `from_config` time and does not model the later replacement pass.** So any peripheral that `configure_cortex_m` installs reads as a permanent stub. `nvic` and `scb` are exactly that, and they are the only two entries in finding (b) I called "not intentional".

With those removed, **(b) collapses to intentional stubs only** — `stub` ×120 and the `nrf54l_*_stub` family ×18. So all three counted paths are cold or benign on the current corpus:

| Path | Result |
|---|---|
| (a) dropped Cortex-M memory errors | 2, both in a fixture built to provoke one |
| (b) stub fallthrough | 163, **all intentional** once the transient is excluded |
| (c) undecoded register access | 5 distinct accesses, no clock-enable offsets |

That does **not** make the underlying code right — firmware that genuinely HardFaults still completes with a green verdict, and that is still worth fixing. It means the *incidence* on shipped labs is near zero, so the fixes are cheap and carry almost no migration risk.

### What should change here before merge

The census should record peripherals **after** machine construction completes, not at factory time — otherwise it will keep producing this false positive for every core-installed peripheral. Until that is fixed, treat finding (b)'s "not intentional" column as unreliable, and do not let the 22+3 figure drive downstream work.

Findings (a) and (c) are unaffected: they instrument runtime access, not construction.

PR #912, opened 2026-08-09
hw(nrf52+l073): live re-captures — oracle couldn't build, two model bugs, 20/20 on L073

w1ne:
Attached the bench board to discharge the `nrf52840` drift_ack. **It did not go clean**, and what it found matters more than the date change.

## 1. The nRF hw-oracle tests had not compiled since 2026-07-18

Seven `nrf52_*` tests call `sim.read_u32`/`write_u32`. The bus-access consolidation removed those inherent `SystemBus` shadows in favour of the `Bus` trait and didn't add the import here. These files build **only** under `--features hw-oracle-nrf52`, which CI never enables because runners have no hardware — so nothing went red.

**The `drift_ack` reading "re-capture pending" was pointing at a path that could not be built, let alone run.** The feature gate that keeps hardware tests off runners is the same gate that hid the rot. That is the manifest's own vacuity hazard, in a spot its header warning doesn't reach.

## 2. SPIM0 PSEL_MISO — sim `0x0`, silicon `0x2E` (mmio was 15/16)

`nrf52840_serial` muxes SPIM0/TWIM0 at one base. PSEL **writes** were already broadcast to both halves — deliberately, with a comment explaining why. PSEL **reads** still dispatched to TWIM, which models only SCL (`0x508`) and SDA (`0x50C`). So SCK and MOSI round-tripped and MISO returned 0.

Not cosmetic: nrfx and Zephyr pinctrl both program PSEL *before* setting ENABLE, so this is on the path real firmware takes.

## 3. SPIM PSEL.CSN (`0x514`) was missing from Nrf52SpiRegs

Not assumed — asked the board. Wrote `0x2B` to `0x40003514`, read `0x2B` back.

## Guard

`serial_instance::psel_block_reads_back_while_disabled` — a plain unit test needing no hardware, so CI catches a regression the hw diff cannot. **Watched failing first**, and it caught `0x514` before I had found that second gap.

## Verification (board attached)

| | |
|---|---|
| hw-oracle nrf52 suites, `NRF52_STRICT=1` | **11 / 11 pass**, mmio now **16/16** |
| `labwired-core --lib` | **2987 passed, 0 failed** |
| Blast radius | **nRF-only** — every `spi.rs` edit is inside `Nrf52SpiRegs`; Stm32/Stm32Fifo/Stm32H5 have their own structs |
| `cargo fmt --check` | clean |
| `generate_validation_status.py --check` | exit 0 |

Two suites initially reported FAILED in 0.00s — that was missing firmware ELFs (`firmware-nrf52840-conformance`, `-cpu-conformance`), not divergence. Built them; both pass.

## Honest scope

- **Same board, not a second one.** `DEVICEID 707dc298` matches the 2026-06-09 baseline. The C3 (#896) and S3 (#898) re-captures were cross-board corroboration; this is a re-read.
- **RNG is still uncovered** by any asserted sweep. That is a *coverage* gap, not a drift gap — re-capturing cannot close it. Both entries now say so explicitly, so `✅ fresh` isn't read as "RNG is validated".
- `seeed-xiao-nrf52840-sense` rides this run by design (same board, same suites) — its entry says so rather than implying an independent capture.

---

## Update — NUCLEO-L073RZ re-captured too, and a correction

**L073RZ: `l0_mmio_diff` 20/20, 0 divergence** under `L073_STRICT=1`. On-board ST-LINK V2.1 `V2J28S17` (serial `066CFF55…5340`), `DBGMCU IDCODE` read back `0x20086447`. Clean on arrival — unlike the nRF, this one found nothing to fix. Scope unchanged: RCC/GPIO/SPI1/TIM2/TIM21, not a full-chip sweep. `nucleo-l073rz` is now `✅ fresh`.

I also checked my own earlier hypothesis that the other feature-gated oracles had rotted the same way as nRF. **They had not** — `--features hw-oracle-stm32` and `--features hw-oracle-c3` both build clean. The rot was nRF-specific.

### Correction: I had been running the wrong gate

Throughout this series I reported "gate exit 0" from `generate_validation_status.py --check`. **That flag only proves the generated doc matches the manifest — it does not test drift at all.** Drift is a separate `--drift` flag, and CI runs both.

Under the real gate, the `spi.rs` edit in this PR **drifted four silicon-tier boards** (`stm32h563`, `nucleo-l476rg`, `stm32f103`, `stm32f407`) — `--check --drift` exited **1**. `spi.rs` is on their watch lists, so their digests moved even though the change cannot affect them.

Fixed properly rather than by re-dating: each is acked with the reason — every edit is inside `Nrf52SpiRegs`, an nRF-only struct, while those boards use `Stm32`/`Stm32Fifo`/`Stm32H5` layouts, separate structs whose executed lines are untouched.

One trap found while doing it: **`--write-ack-digests` re-stamps EVERY ack**, not just the ones you meant. It silently widened `stm32f411ceu6` and `stm32h735` to cover a change nobody had written down. Both are now annotated even though they're sim-validated tier and the gate doesn't fail on them.

Verified: `--check --drift` exits **0** on this tree, and **0** on the `origin/main` baseline — so the drift was mine, and it is covered rather than inherited.


w1ne:
Triaged the conflicts so this is a small decision rather than a large one. I
reproduced the merge locally and did **not** push a resolution — one hunk is
genuinely yours to call. Breakdown of all 8 conflicted files:

**6 files — trivial, take this branch's side.** Every `hw-oracle` test conflicts
on the *same* hunk, and `HEAD` is empty in all six: the branch only adds a
comment.

```
crates/hw-oracle/tests/nrf52_full_register_conformance.rs
crates/hw-oracle/tests/nrf52_gpio_conformance.rs
crates/hw-oracle/tests/nrf52_mmio_diff.rs
crates/hw-oracle/tests/nrf52_onboarding_diff.rs
crates/hw-oracle/tests/nrf52_power_conformance.rs
crates/hw-oracle/tests/nrf52_timer_rtc_conformance.rs
```

**1 file — generated, do not hand-merge.** `docs/boards/VALIDATION_STATUS.md`
(6 conflicts) is rebuilt by `python3 scripts/generate_validation_status.py`.
Resolve the manifest, regenerate, done.

**1 file — needs you.** `validation/manifest.yaml`, 9 conflicts. Eight are
additive commentary and merge by keeping both sides in date order. The two
sides are complementary rather than contradictory, and it is worth being
explicit about why:

* this branch records that a bench board **was** attached on 2026-08-09 and the
  suites re-ran live, discharging the acks that predate it;
* `main` has since gained an ack dated **2026-08-10** for the UARTE `BAUDRATE`
  reset correction (core#924), which is *newer than your capture* and so is not
  covered by it. That one has to survive the merge or the drift gate will fail
  for a real reason.

The hunk I will not decide: one board carries **two competing `drift_ack` +
`drift_ack_digest` values** — yours dated 2026-08-09 (SPIM PSEL, acked without
a live re-capture for that board) versus main's. A board has one ack, so this
is a choice between two hardware claims, and guessing would put evidence in the
manifest that nobody produced.

After resolving: `python3 scripts/generate_validation_status.py` then
`--check --drift` to confirm.

PR #923, opened 2026-08-10
fix(core): C3 ROM-boot analog stimuli

w1ne:
Opening this so the commit can go through core's gates.

`51de3839a` ("Fix C3 ROM-boot analog stimuli") exists only as a branch tip. The
superproject's #1548 pins its gitlink straight at it, which the new
`core pointer does not move backwards` guard rejects — and rightly:

* the product would run engine code that never passed core's main-branch gates;
* the nightly `submodule-bump` pins core's default-branch head, so it would
  silently revert the pin the next morning;
* every other pin bump reads as `diverged` against it, blocking unrelated PRs.

Content is a `crates/core/src/boot/esp32c3_rom.rs` change (+14/−13) with a
`world_esp32c3_ble_pong.rs` test (+39). The branch is 116 commits behind main,
so it needs an update before merge.

Once this lands, #1548 can repin to a commit that is actually on core main.


w1ne:
Closing — empty. The content is already on `main`.

I opened this because superproject #1538/#1548 pin `core` at `51de3839a`, a bare
branch tip with no PR, and the pin guard rejected it. But the fix itself already
landed here by another route:

    crates/core/src/boot/esp32c3_rom.rs
      blob on main        2a2e04bd
      blob on 51de3839a   2a2e04bd     <- identical

After `update-branch` this PR is `+0/-0`.

What is actually broken is on the superproject side: `main` pins `2081bf661`, a
merge commit on `fix/secure-boot-merge-c3-pin` that is **diverged** from core
main (ahead 2, behind 30) — so the product builds its engine from a commit that
is not reachable from core/main, and the nightly `submodule-bump` will revert
it. Both "ahead" commits net to zero content (`compare/main...2081bf661` lists
no changed files), so repointing the superproject at core main head loses
nothing. Doing that separately.

PR #932, opened 2026-08-10
feat(spi): opt-in edge-accurate slave sampling — STM32 bit engine and ESP32-C3 GP-SPI

w1ne:
**This branch has existed since 2026-08-08 with no PR.** Seven commits, ~1,600 lines, pushed and then never opened. Found by sweeping for remote branches with no PR — the same failure recorded twice before in this repo: pushing is not shipping.

Opened as a **draft** because it needs one decision that is not mine to make; see the bottom.

## What it does

A device model can opt into edge-accurate slave sampling (`SpiSampling::Edge`) instead of the frame-boundary byte contract. Honoured by the STM32 classic/FIFO bit engine and the ESP32-C3 GP-SPI, which share **one** edge model (`edge_slave_capture` / `edge_miso_wire`). The controllers with no bit engine — ESP32 classic, ESP32-S3, nRF52 SPIM, STM32H5 SPIv3, Kinetis DSPI — **reject** an opt-in device at config time, naming the controller, rather than ignoring the request silently. A capability the engine cannot honour should fail loudly, not degrade quietly.

## Byte-identical, proven the right way round

The first commit on the branch is `test(spi): record the byte-level wire golden on pristine HEAD` — the golden was captured **before** the feature existed, so it is not a mirror of the new code. `default_byte_level_path_is_byte_identical` still passes after merging 418 commits of main.

## The merge

Four conflicts, all both-sides-additive. Two things a mechanical union got wrong, caught by the compiler rather than by reading:

- `fn sampling()` and `fn needs_external_bus_poll()` were spliced into one body — each side's hunk ended mid-function, and the single closing brace after the conflict could only serve one of them.
- `SpiDevice` was imported twice in `esp32c3/spi.rs`, once from each side.

The `#[cfg(test)]` attributes in `tests/mod.rs` were checked module by module. A union across two lists of `pub mod` declarations is exactly how a dangling attribute silently re-gates the next module; every module carries the gate main gives it.

## Verification, by exit code

| Check | Exit | Result |
|---|---|---|
| `spi_byte_level_golden` | 0 | byte-identical default path |
| `labwired-core --lib` | 0 | 3070 passed |
| `firmware_survival` | 0 | 61 passed |
| `labwired-hw-oracle` (offline) | 0 | all silicon-diff suites green |
| `cargo fmt --all --check` | 0 | after rustfmt reordered the merged mod list |

## ⚠️ The decision this draft is waiting on

The drift gate fails at the committed tree, **exit 1**, for five boards: `esp32c3`, `nucleo-l476rg`, `nucleo-l073rz`, `stm32f103`, `stm32f407`. They watch `peripherals/spi.rs` and `esp32c3/spi.rs`, and this branch genuinely changes both.

**I have not written a drift_ack.** An ack is a human assertion about silicon, and unlike the digest-only churn acked earlier today, this one changes engine code rather than just moving a source digest. What can be said from evidence:

- the byte-level golden is unchanged, so a device that did not opt in sees the same wire it saw before;
- every one of those five boards' own offline silicon-diff suites passes against its frozen capture on this branch (`esp32c3_reset_conformance`, `l476_mmio_diff`, `f103_conformance`, `stm32f1_mmio_diff`, `stm32f4_mmio_diff`, `stm32l0_mmio_diff`, all exit 0).

That is a strong argument that nothing observable moved for these boards, but it is an argument, not a capture. Mark it ready when you have either acked the five or re-captured them.


w1ne:
`pr-gate` is red only on the regen + drift gate. Triage so this is quick:

```
ERROR: silicon validation has DRIFTED (model changed after last capture,
no covering drift_ack):
  esp32c3
  nucleo-l476rg
  nucleo-l073rz
  stm32f103
  stm32f407
```

Exactly the chips this PR's SPI work touches — expected, not a surprise.

The ack should be easy to write honestly here, because the PR already carries
its own evidence:

* `declarative_spi.rs` doc: *"Opt in to edge-accurate (bit-level) slave sampling
  in this SPI mode (0..=3). **Omit for the default byte-level frame exchange.**"*
* the behaviour is gated on `opt_in: bool` (`if self.opt_in { … }`), default off
* `crates/core/src/tests/spi_byte_level_golden.rs` (+189) exists precisely to
  pin the default path

So the reason writes itself: opt-in, default off, default byte-level exchange
pinned byte-identical by the golden test — no live re-capture claimed. That is
a much stronger position than most drift acks get to stand on.

```
python3 scripts/generate_validation_status.py --write-ack-digests
# add the per-board reason above each drift_ack
python3 scripts/generate_validation_status.py            # regenerate the doc
python3 scripts/generate_validation_status.py --check --drift   # expect exit 0
```

Watch for `--write-ack-digests` stamping **more** boards than the five listed:
it re-stamps anything whose watched files moved, and re-stamping without a
reason is how this gate decays. On #924 it silently caught four extra boards
that needed their own note.

I have not pushed anything — the fidelity claim for a 1627-line SPI engine
change belongs to whoever can vouch for the golden test, not to me.

w1ne:
Noticed you moved this to draft — leaving it alone. Two commits from me are on
the branch that you have not pulled yet (your worktree is still at `43835998`),
so here is what they are, in case they are useful or in case you want them gone.

**`647c35f7` — drift acks.** The gate named five boards; `--write-ack-digests`
re-stamped **eight**, the extra three being `stm32h563`, `stm32h735`,
`stm32f411ceu6` whose watched files also moved. All eight got a written reason
rather than a silent re-stamp.

I did not take the opt-in claim on trust — I ran it on the branch:

```
cargo test -p labwired-core --lib spi_byte_level_golden
tests::spi_byte_level_golden::default_byte_level_path_is_byte_identical ... ok
```

The acks cite that, and they also state what they do NOT cover: once a lab opts
in, its timing is outside every capture these boards have.

**`4bac8122` — clippy.** Four `-D warnings` errors in
`spi_edge_sampling_lab.rs`: three `doc list item overindented` (continuation
lines aligned under the arrow instead of the list content) and one
`type_complexity` on the cases tuple. I named the factory type rather than
adding `#[allow]`, since the boxed factory is the point of that guard and an
allow would have hidden the reason.

`clippy --lib --tests -- -D warnings` 0 errors, `fmt --check` clean,
`spi_edge_sampling_lab` 7 passed, and the PR went CLEAN before you drafted it.

Drop either commit if you would rather write them yourself — I will not touch
the branch again while it is a draft.

Issue #928, opened 2026-08-10
world_esp32c3_ble_pong: host election is flaky (~2 in 3) on the committed fixture


w1ne:
`two_c3_nodes_running_one_ble_pong_image_start_the_game` fails about two runs in
three on `main` with the **committed, unmodified** flash fixture.

```
assertion `left == right` failed: exactly one node must elect HOST
  crates/core/tests/world_esp32c3_ble_pong.rs:366
```

## Reproduction

`dcf65286`, release + `event-scheduler`, fixture
`esp32c3-ble-pong-flash.bin` at its committed hash `f3cc324f…`, four
consecutive runs of the same binary:

```
run 0  FAILED    ...start_the_game
run 1  ok        3 passed
run 2  FAILED    ...start_the_game
run 3  FAILED    ...start_the_game
```

The other two cases passed every time, including
`two_c3_nodes_in_one_lab_do_not_share_a_bluetooth_address`, so the ROLE banner
and BLE addressing are stable. Only the election flakes.

## Why it is a race

Both nodes start unstrapped, so each must hear the other's advertisement and
compare tags before either can lock a role:

```cpp
if (!roleLocked) {
  if (!strapHost) isHost = (myTag < peerTag);
  roleLocked = true;
```

`run_pong` advances the two nodes in alternating fixed slices rather than
interleaving them, so whether an advertisement lands inside the peer's scan
window depends on where the slice boundary falls. Nothing in the harness pins
that relationship, and the cycle budget is the only thing bounding it.

## Why it matters

This test is the repo's claim that "a real user's real sketch works" for a
published lab. A gate that fails two runs in three teaches everyone to re-run
it, which is how a real regression gets waved through — and this one is
release-only on push-to-main, so the noise lands where nobody is watching a PR.

## Not the same as #927

Discovered while running #927's control, but distinct: this reproduces on the
CURRENT committed fixture and the console is fully populated. #927 is the
refreshed image producing NO console output at all. Please keep them separate —
fixing the flake will not make #927 pass, and vice versa.

Issue #927, opened 2026-08-10
esp32c3 twin: a layout-only rebuild of the BLE Pong sketch boots to an empty console — fixture refresh blocked, core-integrity red


w1ne:
> **Rewritten 2026-08-10.** The original body offered "engine gap **or** broken
> lab" and named the shifted DRAM segment as the likely suspect. Four
> hypotheses have since been eliminated and the trigger bisected to a single
> source line; the evidence now points at the twin, not the sketch. The
> superseded reasoning is preserved in the thread below.

`core-integrity` is red on every `main` commit because
`scripts/ci/check-published-lab-drift.sh` sees the owner's published project
`c477f82961e86f601e7b908ae7e12311` has moved past the committed fixture. The
refresh the failure message asks for cannot land: **the current published sketch
produces no serial output at all in the twin.**

Still red at `466cef2` — run `31397237311`, where the *only* failing step in
`core-integrity` is `BLE Pong fixture drift (committed .ino ↔ LIVE published lab)`.

### The drift is two cosmetic lines

```diff
-  oled.setTextSize(1);
+  oled.setTextSize(2);
-  oled.setCursor(W / 2 - 12, 0);
+  oled.setCursor(W / 2 - 18, 0);
```

Nothing else. Scoreboard font size and its x offset.

### The trigger is `setTextSize(2)`, alone

Rebuilt with only that one line changed, keeping the original cursor: fails
identically (0 passed / 3 failed, empty console). The cursor edit is irrelevant.

| build | app bytes | entry |
| --- | --- | --- |
| old, works | 1049504 | `0x403823f6` |
| `setTextSize(2)` only, fails | 1035104 | `0x4038246c` |
| both lines, fails | 1035104 | `0x4038246c` |

`setTextSize(2)` **shrinks** the image by 14400 bytes and moves every segment.
The two failing builds are byte-identical in size and entry — they differ by one
immediate operand.

### Why this reads as the engine, not the lab

`setTextSize` is called inside `draw()`. `draw()` is called from `loop()`, which
runs only after `setup()` has already executed `Serial.print("ROLE ")`. There is
no order of execution in which that line prevents the banner that precedes it.
Yet the banner is absent, on a CPU that steps ~9M times with zero errors and
zero undecoded instructions.

Control on the same build with the committed image: ROLE banner present in 3 of
3 runs. Old image always prints it, new image never does.

### Eliminated, do not re-spend time here

| # | hypothesis | why not |
| --- | --- | --- |
[...]

PR #938, opened 2026-08-10
ci(integrity): a live monitor can no longer hide the engine's own gates

w1ne:
## The hole

`core-integrity` runs the BLE Pong **published-lab drift check** — a network read against production — immediately before four steps that gate the engine itself. In GitHub Actions a failed step skips the rest of the job. So a third-party monitor sat in front of this repo's own gates.

It fired:

| When | What |
|---|---|
| 2026-08-09 15:53Z | drift check goes red — the owner edited the published sketch (three lines of OLED text size) |
| 2026-08-09 16:39Z | **#908 merges and breaks `world_esp32c3_ble_pong`** |
| …for the next 29 hours | that test never ran again — not on #908, not on any merge after it |

Two ESP32-C3 nodes stopped electing a host, so **the published two-node lab does not start a game**, and every merge since has been green on `pr-gate` while main carried it.

## Evidence for #908

Same fixture, same command, bisected:

| Commit | Result |
|---|---|
| `11207ae1` last green | pass |
| `a67876bd` parent of the suspect | pass |
| `bd32b101` — #908, "Fix C3 ROM-boot analog stimuli" | **FAIL** — `left: 0, right: 1`, "exactly one node must elect HOST" |

```
nodeA tag=5 G ball=72,36 v=2,1 me=0 peer=0 score=0:0
nodeB tag=6 G ball=72,36 v=2,1 me=0 peer=0 score=0:0
```

Not a cycle budget: at 3× `PONG_SOAK_CYCLES` the ball reaches 123,61 and neither node ever sees a peer. #908's only engine change is swapping the ROM-boot SAR ADC stub for the shared behavioural controller.

## What this PR changes

The masking, and only the masking. The four engine gates after the monitor carry `if: !cancelled()`, so they run and report regardless of what the network-facing step did. Both still fail the job; neither can stand in front of the other.

## What it deliberately does not change

**#908 itself.** A straight revert restores the two-node lab — proven, its parent passes — but turns #908's own `shipped_pong_firmware_observes_gp2y_distance_through_gpio1_adc` red, because that test exists to prove the new ADC path observes a live stimulus. That is a choice about which user-facing behaviour matters more, not a mechanical revert, and the mechanism linking an ADC swap to a dead BLE election is not yet established.

**The fixture refresh.** I rebuilt it (live sketch, hosted recompile, four blobs verified by digest, merged at `0 / 0x8000 / 0xe000 / 0x10000`) and the drift check goes green with it. It is not here because **the recompiled image never reaches its serial banner in the twin** — dead at the green parent commit too, so it is a separate defect from #908. Same IDF `v4.4.7`, byte-identical bootloader, same five segments at the same load addresses, all the sketch's own strings present, app 14,400 bytes smaller. Landing that fixture would pin an image that does not boot.

### DIFF
--- a/.github/workflows/core-ci.yml
+++ b/.github/workflows/core-ci.yml
@@ -867,6 +867,16 @@ jobs:
       - name: BLE Pong fixture drift (committed .ino ↔ LIVE published lab)
         run: scripts/ci/check-published-lab-drift.sh
 
+      # ⚠ EVERY ENGINE GATE BELOW CARRIES `if: !cancelled()`, AND THE STEP ABOVE
+      # IS WHY. The drift check reads production over the network, so it fails
+      # for reasons that have nothing to do with the commit — and a failed step
+      # skips the rest of the job by default. On 2026-08-09 it went red at 15:53
+      # because the owner edited the published sketch. #908 merged 46 minutes
+      # later and broke `world_esp32c3_ble_pong` (two C3 nodes stopped electing
+      # a host); that test sits in the step below and DID NOT RUN for the next
+      # 29 hours, on that merge or any after it. A third-party monitor must
+      # never be able to stand in front of this repo's own gates.
+
       # `esp32c3_usb_cdc_console` is here for COST, not because it is
       # release-only: same reason as `world_esp32c3_ble_pong` above. It boots
       # the C3 mask ROM, which is 335s in debug (measured) and seconds in
@@ -881,6 +891,7 @@ jobs:
       # its own headline behaviour — CDC-on-boot builds actually printing — with
       # no lane running the test that proves it.
       - name: Release-gated tests — cfg(not(debug_assertions))
+        if: ${{ !cancelled() }}
         run: >
           cargo test --release -p labwired-core --features event-scheduler
           --lib --test event_scheduler --test release_only_cfg_ratchet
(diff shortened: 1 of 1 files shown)


w1ne:
### Pre-merge verification: reproduced the masked failure, and what merging this will look like

Ran the masked test locally against the **committed, unmodified** fixture (`f3cc324f…`) at current core `main` `6d946d08`, release + `event-scheduler`, four consecutive runs of the same binary:

```
RUN 1  FAILED   assertion `left == right` failed: exactly one node must elect HOST
RUN 2  FAILED   exactly one node must elect HOST
RUN 3  FAILED   exactly one node must elect HOST
RUN 4  ok       3 passed
```

3 of 4. The other two cases passed every run. So the thing this PR unmasks is real and still present on `main` today — it is not historical.

One nuance worth recording, since #928 and this PR describe it slightly differently: it is **not deterministic**. This PR's bisect table reads as a clean pass/fail per commit, and #928 calls it flaky ~2 in 3. At 3-in-4 here, both are consistent with a high-rate race rather than a hard break — which matters for the bisect, because a single passing run on a candidate commit is not enough to clear it. `a67876bd`'s "pass" would want a few repeats before #908 is settled as the trigger.

### What to expect immediately after merging

`core-integrity` will get **worse-looking**, and that is the fix working. Today it fails at the drift step and skips everything after. Once these four steps carry `if: !cancelled()`, the release-gated step actually runs and will fail roughly three runs in four on `world_esp32c3_ble_pong`.

Flagging it so the new red is not read as this PR breaking something. It is 29 hours of accumulated breakage becoming visible for the first time.

### Not to be confused with #927

Both issues name this test, and they are different failures:

| | fixture | fails at | symptom |
|---|---|---|---|
| this / #928 | committed `f3cc324f…` | `:366` | election — one node never sees a peer |
| #927 | rebuilt from live sketch | `:241`/`:256` | no console at all, all 3 cases |

I hit #927's variant earlier today attempting the fixture refresh. Fixing either one leaves the other.

PR #947, opened 2026-08-11
docs: drop llms.txt from site navigation

w1ne:
## Summary
- Remove **llms.txt (agents)** from `mkdocs.yml` nav. Page file stays available if linked directly.

## Test plan
- [ ] Nav no longer shows llms.txt after deploy


w1ne:
Already shipped in #948 (llms dropped from nav with sidebar change).

PR #973, opened 2026-08-14
chore(chips): promote esp32c3, and say what stm32l073 actually needs

w1ne:
The chips gate did the thing it was built for. Run against the freshly released v0.22.0 it did not quietly stay green on a stale note — it **failed**:

```
esp32c3  blocked  FIXED — promote it
::error::esp32c3 is marked blocked but now runs. Move it to "how": "asset".
```

v0.22.0 carries the fix for `UART type 'esp32c3_uart' has no register layout modelled yet`, so the C3 boots its demo:

```
C3 BLINKY BOOT
LED ON
LED OFF
```

It is an `asset` entry now, asserted on that first line.

**stm32l073 changed reason, not state.** On v0.21.0 the CLI could not parse the descriptor we ship for it (`peripherals[34].clock: invalid type: sequence`). v0.22.0 reads it fine and the run faults instead — the firmware it was paired with is built for Cortex-M4 and this part is M0+. So it is not blocked by the CLI, it needs firmware: `firmware-l073-demo` builds in this repo and only needs publishing through the demo-firmware workflow (#968).

**17 of 26** documented chips now run from a plain install with no toolchain, and **nothing is blocked by the released CLI**.

### DIFF
--- a/scripts/ci/docs-runnable-chips.json
+++ b/scripts/ci/docs-runnable-chips.json
@@ -29,11 +29,11 @@
       "expect": "faithful ROM loaded"
     },
     "esp32c3": {
-      "how": "blocked",
+      "how": "asset",
       "firmware": "demo-esp32c3-blink.elf",
       "sha256": "aea04c8527eff25e8969b851dc9cc2436cf13e40c6019788741b7eecd5004789",
-      "expect": null,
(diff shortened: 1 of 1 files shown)


w1ne:
Folded into #972 — both are release follow-up housekeeping, and the shared runners were making each merge cost the next PR a full CI cycle.

PR #976, opened 2026-08-14
ci(chips): gate that every chip is executed by a CLI gate

w1ne:
See commit message: the gate derives per-chip CLI coverage from what CI actually runs (Arduino matrix cells the workflow really names, plus scripts a workflow really passes to `labwired test`), rather than from any list of its own.

**The gap it closes.** The perf gate answers *is this chip measured* from disk. Nothing answered the same question for behaviour — both matrices name their cells inline, so a chip could land in `configs/chips/` and have no gate run firmware on it. 8 of 26 were in that state.

**What deriving it found.** Four smokes already in the tree, passing, run by no workflow: `nrf54l15-dk` (4/4, 0.05s), `nrf54l15-smart-ring` (7/7, 0.09s), `stm32f411ceu6-blackpill` (10/10, 2.4s), `kw41z-cow-activity` (1/1, 2.3s). All run committed ELFs, so wiring them into the coverage matrix costs seconds and no toolchain. CLI coverage goes 18/26 → 21/26; `nrf54l15`, `stm32f411ceu6` and `mkw41z4` get their first CLI proof.

**The ledger.** `configs/ci/chip-coverage.yaml` holds only the remaining five (`esp32s3-zero`, `nrf5340`, `rp2350`, `stm32f405`, `stm32f767`), each with a reason and a kind. `max_uncovered` must equal the entry count, so closing a gap forces the ceiling down — the shrink-only contract from `KNOWN_GAPS`.

**Non-vacuity.** Four failure modes were driven against the tree and each reddens the gate: a new chip with no cell; a board dropped from the workflow matrix but left in `boards.yaml`; a script removed from the coverage matrix; a ledger row kept after its chip became covered. The 13 pytest cases grade synthetic trees via `--repo-root`, so they cannot pass merely because today's tree is in the right state.

**Not in scope:** the browser lane — that join is only visible in the superproject.


w1ne:
Read this against `scripts/ci/docs-runnable-chips.{json,sh}`, which landed today and enumerates the same 26 chips. **They should both survive, but not both be called "chip coverage"** — and the naming is currently close enough to guarantee confusion.

## They answer different questions

| | #976 | docs-runnable-chips |
|---|---|---|
| Question | does **our CI** execute firmware on this chip? | can **a user** run this chip from a clean clone with the released CLI? |
| Oracle | what the workflows actually run | the installed binary, in the canary |
| Coverage is | **derived** — cannot be claimed | **declared**, then executed |
| Runs on | pull requests | every 6 h, macOS + Ubuntu |
| Verdict today | 21/26, 5 waived | 26/26 |

Neither number is wrong, and that is exactly the problem: a reader who sees both concludes one gate is lying.

Each catches something the other cannot:

- **#976 catches** a chip whose model regresses between canary runs, and firmware sitting in the tree that no workflow executes — it found four such smokes, which is real value my gate does not provide.
- **Mine catches** a chip that CI exercises happily with the workspace build while being impossible for anyone who installed the CLI. That is not hypothetical: `esp32c3` was green on main and refused the board on v0.21.0 for two weeks, and `configs/chips/stm32l073.yaml` used a schema the released binary could not parse.

## The disagreement is informative, and worth keeping visible

The five chips this ledger waives are precisely the ones my map proves with a **sibling part's** firmware or a generic image:

```
esp32s3-zero   tier1/esp32s3.elf      board variant, same image
stm32f405      tier1/stm32f407.elf    F4 family image, not part-specific
stm32f767      tier1/stm32f407.elf    no F7 fixture exists
nrf5340        nrf5340-zephyr-hello   upstream Zephyr, boots
rp2350         demo-rp2350-smoke.elf  published today
```

So the honest joint statement is: *26 of 26 run for a user; 21 of 26 are exercised by a part-specific gate in CI.* Both true, and the gap is a real backlog rather than a contradiction.

## Concrete suggestions

1. **Rename to the question, not the mechanism.** "CI chip coverage" here; "documented chips runnable from an install" for mine. The current title reads as the same claim.
2. **One cross-check instead of two lists.** Any chip my map marks `needs-build` or `blocked` must also appear in this ledger — a chip a user cannot run is certainly not covered end to end. Trivially satisfied today (my list is empty), which is the right time to add it.
3. **Say the delta out loud** in `chip-coverage.yaml`: for `stm32f405` and `stm32f767`, "user-runnable via the F4 family image; waived here because no part-specific fixture exists" is a stronger reason than the current one, and it points at the work.

I have not touched this PR. If you want, I will do (2) as a one-file follow-up once this lands.

PR #978, opened 2026-08-14
feat: close the loops the install work left open

w1ne:
Four gaps this session named and did not fix. All four are the same shape: something true about the product that nothing in the repo could observe.

## A release that stops happening is invisible

Every suite here runs against the working tree, which is by definition current. So nothing could see that main was **984 commits and 18 days** ahead of the released CLI — and three separate "product defects" turned out to be exactly that gap:

- examples using script fields the released CLI could not parse,
- `esp32c3` refusing a board it runs perfectly on main,
- `configs/chips/stm32l073.yaml` rejected outright by the released binary.

`scripts/ci/release-freshness.sh` measures it against the tags (21 days / 250 commits) and runs in the **canary, not on pull requests**: a PR did not cause the drift and cannot fix it, so failing one would be blaming the wrong change.

## The CLI could not report a crash

The playground got a reporter; the surface that runs on the widest range of machines stayed silent. A panic hook — which still runs under `panic = "abort"` — posts the same enumerated row the installer does. Never the panic message: it routinely carries a file path or a customer's chip name. It chains to the previous hook, so the backtrace the user sees is unchanged. Off with `LABWIRED_TELEMETRY=0`, `DO_NOT_TRACK`, or `CI`.

## The chips gate had never been seen to fail

Its first version would have passed a chip whose run printed nothing and faulted silently — which is what a Cortex-M4 image does on an M0+ part. `test-docs-runnable-chips.sh` now hands it three broken maps and requires a red from each:

```
ok    a chip paired with another part's firmware fails
ok    a chip missing from the map fails
ok    an expectation the firmware never prints fails
ok    a correct map passes
```

That last line matters as much as the others: without it the three above would also be satisfied by a gate that fails at everything. Cut-down maps keep the whole thing at five seconds.

## The extension is on no registry

Marketplace 404, Open VSX 404, and the only install path is a hand-downloaded `.vsix` with **0 downloads**. The canary now checks both and fails while that is true, with the exact next step in the error message. It goes green the day the publisher and namespace exist — no code change needed.


w1ne:
Folded into #979, which was branched on top of this one and carries every commit here — merging both would have cost an extra full CI cycle on two shared runners for no additional change.

PR #990, opened 2026-08-14
feat(ci): rp2350 runs in a gate — the chip-coverage ledger is now empty

w1ne:
The last entry in `configs/ci/chip-coverage.yaml`. `crates/firmware-rp2350-demo` built and `rp2350_clkrst_profile.rs` pinned the reset path, but the chip had **no system manifest**, so no script could name a firmware and a board together and no workflow executed RP2350 at all. This writes the two artifacts that were missing — `examples/rp2350-demo/{system,io-smoke}.yaml` — and wires the cell. 2/2, 0.05s.

**What it proves that the reset-path test does not:** a `cortex-m-rt` image boots from the XIP flash base `0x1000_0000`, reaches `main`, and its writes reach the PL011 at the RP2350 UART0 base `0x4007_0000` — which is *not* the RP2040 base.

**Non-vacuity checked, not assumed:** on the `rp2040` descriptor (UART0 at `0x4003_4000`) the same script captures **zero** UART bytes and drops to 1/2.

**`build_dir` is load-bearing.** The linker script comes from the crate's own `.cargo/config.toml` (`-C link-arg=-Tlink.x`), and cargo reads config from the *working directory's* ancestors, not from the package it is building. Measured both ways on this tree — the same `cargo build -p firmware-rp2350-demo --release --target thumbv8m.main-none-eabi` gives:

| cwd | entry |
|---|---|
| `crates/firmware-rp2350-demo` | `0x10000801` |
| workspace root | `0x00000000` |

A 0x0 image boots nowhere and prints nothing, so the smoke would fail with an empty UART and no clue why. The new guard step names that failure and says which cell field fixes it. `build_dir` defaults to `.`, so every existing cell is unchanged.

**Ledger: 26 of 26 chips proven by a CLI gate, `uncovered: {}`, ceiling 0.** That ceiling is the point — a new descriptor now reddens the gate on the PR that adds it, and the only way to green it is a script and a cell that run the chip. 8 → 5 → 3 → 2 → 1 → 0, never the other way.


w1ne:
pr-gate caught something real, and the fix is in `4cae7ad3`.

The gate's own test says it plainly: *"if nothing is uncovered, drop the ledger rather than keeping an empty one"*. Closing rp2350 made that true, and my `uncovered: {}` was exactly the state it forbids.

`configs/ci/chip-coverage.yaml` is now **deleted**. A missing ledger means no declared gaps, which is **stricter** than an empty one, not looser: `evaluate` calls every uncovered chip an error, so with no file present a new descriptor reddens the gate on the PR that adds it. Deleting the file is not a way around the gate — it is the gate at full strength.

Two new tests pin that, because "fails closed" is a claim, not a comment:
- an uncovered chip with **no** ledger file → exit 1, and the output names the chip;
- a fully covered tree with no ledger file → exit 0.

The committed-tree test now asserts the invariant both ways: the file exists exactly while there is something to declare. Full perf-gate suite locally: **80 passed**.

Issue #961, opened 2026-08-12
cpu_trace_conformance: Avr core is not exercised by the standardized trace suite


w1ne:
Surfaced by the new PR workspace-test lane (#958): `labwired-core --test cpu_trace_conformance::every_cpu_core_is_covered_by_this_file` is RED on main.

```
assertion `left == right` failed: cores implementing `Cpu` and cores exercised by this file have diverged.
  left:  ["Avr", "CortexM", "RiscV", "XtensaLx7"]
  right: ["CortexM", "RiscV", "XtensaLx7"]
crates/core/tests/cpu_trace_conformance.rs:344
```

The ATmega328P / AVR Nano twin landed in #940 (and #951) without adding the `Avr` core to `cores()` in `crates/core/tests/cpu_trace_conformance.rs`. The suite exists to force exactly this: every core that implements `Cpu` must emit the standardized instruction trace.

Why nothing caught it: `cargo test --workspace` ran only in nightly `core-full`, which has been failing earlier (at `environment_runner`, #929) since 2026-08-10 and never reaches labwired-core's targets. Verified independently on pristine origin/main.

Fix: add `Avr` to `cores()` and make the AVR core emit the standardized trace (or argue in the file why it is exempt). Until then it is allow-listed in `scripts/ci/workspace-test-shards.json` as known-red with this issue number.

w1ne:
Resolved on `main` by 49f215f6 (`fix(ci): make the AVR core traceable and three red gates green`): `cpu_trace_conformance` now includes and exercises the AVR core.

PR #991, opened 2026-08-15
fix(core): the Cpu trait stops fabricating runtime snapshots

w1ne:
Ledger row 4.5.

## What was wrong

`crates/core/src/lib.rs`, the `Cpu` trait:

```rust
fn runtime_snapshot(&self) -> (runtime_snapshot::CpuKind, Vec<u8>) {
    (runtime_snapshot::CpuKind::ArmCortexM, Vec::new())
}

fn apply_runtime_snapshot(&mut self, _kind: CpuKind, _bytes: &[u8]) -> SimResult<()> {
    Ok(())
}
```

Two cores hit these defaults today: **AVR**, which reported its arch as
`ArmCortexM` — a `CpuKind` that does not even have an `Avr` variant — and
**Cortex-M**, which reported a well-formed `ArmCortexM` snapshot with an
**empty body**. `snapshot capture` wrote that to a file;
`WasmSimulator::take_runtime_snapshot` handed it to JS. Nothing downstream
could tell either apart from a real capture.

The restore half was worse: it discarded the bytes and returned `Ok(())`, so
"restored" and "silently ignored" were the same observable outcome. In the
browser that is a reported-good resume onto a machine with warm peripherals
and a cold CPU — a machine that has never existed on silicon.

A prior PR added `supports_runtime_snapshot()` and a doc comment explaining
why the fabrication was there. That mitigated the three call sites that
remembered to ask, and left the fabrication itself in place.

## What changed

The wrong answer is now unrepresentable rather than checked for — the same
approach `system::arch_policy` takes to refusing to guess an architecture.

* `Cpu::runtime_snapshot` returns `Option<(CpuKind, Vec<u8>)>`. `None` is the
  only way to say "I cannot"; the type carries no arch tag to invent.
* `Machine::take_runtime_snapshot` returns `Option<MachineRuntimeSnapshot>`
  for the same reason — a machine snapshot missing its CPU half is not a
  cheaper snapshot, it is a resume that starts from a cold core.
* `Cpu::apply_runtime_snapshot`'s default is
  `Err(SimulationError::NotImplemented)`. The CPU half is applied **first** in
  `Machine::apply_runtime_snapshot`, so a refusal fails the call before a
  single peripheral is touched.
* `supports_runtime_snapshot` is removed. It was a second source of truth for
  the same fact and could disagree with what a capture actually contained. All
  three callers (`snapshot capture`, `--capture-app-entry`,
  `WasmSimulator::take_runtime_snapshot`) gated on it and then immediately
  took the snapshot, so they now just handle the `None`.

## Existing chips are untouched

RISC-V and Xtensa LX7 are the only cores that model runtime snapshots. Their
blob bytes, `CpuKind` and restore paths are unchanged — only the wrapper moved
from `(k, v)` to `Some((k, v))`. `RUNTIME_SNAPSHOT_VERSION` is untouched and no
committed artifact moves. Both CLI resume paths are RISC-V-only and keep
working; the ESP32-C3 resume tests are unchanged apart from the `Option`.

## Tests

New `crates/core/tests/cpu_runtime_snapshot_honesty.rs` (11 tests):

* an ISA with a real snapshot round-trips — Xtensa LX7, RISC-V, and through
  the `Box<dyn Cpu>` forwarding impl the browser runtime holds;
* an ISA without one reports `None` instead of claiming Cortex-M — Cortex-M
  itself and AVR;
* a refused restore is a different `Result` than a successful one, at both the
  CPU and the `Machine` level, and a refused `Machine` restore leaves the PC
  where it was;
* a source scan holding the two halves together: a core implements capture and
  restore, or neither, never one. The scan asserts it *found* the cores that do
  implement them, so a pattern that matches nothing fails rather than passing
  vacuously.

### Negative control

The behavioural assertions were first written against the **old** signatures
and watched failing on pristine `origin/main` (commit 22574f888):

```
test result: FAILED. 1 passed; 6 failed
  negctl_avr_never_reports_itself_as_cortex_m
  negctl_avr_restore_of_a_foreign_blob_is_not_ok
  negctl_cortex_m_does_not_claim_an_empty_snapshot_is_a_snapshot
  negctl_discarded_restore_is_distinguishable_from_a_successful_one
  negctl_machine_restore_onto_a_cpu_that_cannot_restore_is_not_ok
  negctl_machine_without_a_cpu_snapshot_refuses_to_produce_one
```

The one that passed was the Xtensa round-trip — the control proving this PR
does not change the ISAs that already work.

The pairing ratchet was driven red in both directions by temporarily giving
`Avr` one half: restore-only (`left: [riscv, xtensa]` vs
`right: [avr, riscv, xtensa]`) and capture-only (the mirror image, plus
`avr_never_reports_itself_as_cortex_m` going red).

## Local runs

| command | pristine `origin/main` | this branch |
| --- | --- | --- |
| `cargo check --workspace --all-targets` | 0 | 0 |
| `cargo clippy --all-targets -- -D warnings` (CI's default-members lane) | — | 0 |
| `cargo clippy -p labwired-wasm --all-targets -- -D warnings` | — | 0 |
| `cargo test -p labwired-core --lib` | 3092 passed, 0 failed | 3092 passed, 0 failed |
| `cargo test -p labwired-core --test runtime_snapshot` | 5 passed, 1 ignored | 5 passed, 1 ignored |
| `cargo test -p labwired-core --test e2e_esp32c3_snapshot_resume` | 1 ignored | 1 ignored |
| `cargo test -p labwired-core --test cpu_runtime_snapshot_honesty` | n/a | 11 passed |
| `cargo test -p labwired-cli --lib` | — | 0 |
| `cargo test -p labwired-wasm --lib` | — | 43 passed, 7 ignored |

**Pre-existing red, not caused here:** `cpu_trace_conformance ::
every_cpu_core_is_covered_by_this_file` fails identically before and after
(2 passed, 1 failed) — the `known_red` entry for #961.

`cpu_runtime_snapshot_honesty` lands in PR shard 2 automatically
(`workspace_test_shard.py plan` confirms), so no shard-config change is needed.

### DIFF
--- a/crates/cli/src/commands/snapshot.rs
+++ b/crates/cli/src/commands/snapshot.rs
@@ -498,22 +498,18 @@ pub(crate) fn run_snapshot_capture(
 
     dump_trace(&ring);
 
-    // Ask before taking. `Cpu::runtime_snapshot` no longer panics on the arches
-    // that do not implement it (a panic in wasm is a trap, and a trap leaks
-    // wasm-bindgen's borrow guard — see the note on the trait), so an ungated
-    // call here would no longer fail loudly: it would write a well-formed file
-    // whose CPU blob is EMPTY. A resume from that blob restores no registers at
-    // all, and nothing downstream can tell it apart from a real capture. Refuse
-    // instead — a missing snapshot is recoverable, a lying one is not.
-    if !machine.cpu.supports_runtime_snapshot() {
+    // `None` means this CPU models no runtime snapshot. There is no longer a
+    // separate capability query to ask first: the take itself is the answer,
+    // so the check cannot disagree with what a capture would contain. Refuse —
+    // a missing snapshot is recoverable, a file whose CPU blob is empty but
+    // which looks like a real capture is not.
+    let Some(snap) = machine.take_runtime_snapshot() else {
         eprintln!(
             "error: this CPU has no runtime-snapshot implementation, so no resumable \
              snapshot can be captured for it (supported: RISC-V, Xtensa LX7)."
         );
         return ExitCode::from(EXIT_RUNTIME_ERROR);
-    }
-
-    let snap = machine.take_runtime_snapshot();
+    };
     let bytes = snap.to_bytes();
 
     if let Some(parent) = args.output.parent() {
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -2564,20 +2564,14 @@ fn execute_test_loop<C: labwired_core::Cpu>(
             let pc = machine.cpu.get_pc();
             let reached = cap.target_pc == Some(pc) || (0x4200_0000..0x4400_0000).contains(&pc);
             if reached {
-                // Same reason as `snapshot capture`: the trait default is now a
-                // non-panicking EMPTY blob, so an ungated call would write a
-                // resume file that restores no CPU state and still looks valid.
-                // Say so and write nothing. NOT a `continue` — the rest of this
-                // loop body is what actually advances the machine, so skipping
-                // it would hang the run instead of just declining the capture.
-                if !machine.cpu.supports_runtime_snapshot() {
-                    error!(
-                        "capture-app-entry: this CPU has no runtime-snapshot implementation \
-                         (supported: RISC-V, Xtensa LX7) — no snapshot written to {:?}",
-                        cap.path
-                    );
-                } else {
-                    let mut snap = machine.take_runtime_snapshot();
+                // Same reason as `snapshot capture`: a CPU that models no
+                // runtime snapshot answers `None`, and writing a resume file
+                // without a CPU half would produce something that still looks
+                // valid. Say so and write nothing. NOT a `continue` — the rest
+                // of this loop body is what actually advances the machine, so
+                // skipping it would hang the run instead of just declining the
+                // capture.
+                if let Some(mut snap) = machine.take_runtime_snapshot() {
                     snap.set_self_key(cap.chip, cap.fw_sha);
                     if let Some(parent) = cap.path.parent() {
                         let _ = std::fs::create_dir_all(parent);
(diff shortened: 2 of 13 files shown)


w1ne:
## pr-gate is red on the esp32c3 drift gate — deliberately NOT re-stamped

`pr-gate` step **"Validation status — regen check + drift gate"** fails:

```
  esp32c3
```

Derived, not guessed:

* pristine `origin/main` (22574f888) passes the same gate locally — exit 0. This
  PR causes it.
* Two of the files this PR edits are in esp32c3's `models` watch list in
  `validation/manifest.yaml`: `crates/core/src/cpu/riscv.rs` and
  `crates/core/src/peripherals/esp32c3/gpio.rs`. Their newest commit date is now
  2026-08-15, past the `drift_ack: 2026-08-13`.
* The generated diff is the source digest and the status text only —
  `f364ebc696e57dae` → `a79a88552908426a`, and the row flipping from
  "⚠ drift acked 2026-08-13" to "✖ DRIFT". No silicon capture, no golden, no
  reset-value table moves.

What actually changed in those two files:

* `riscv.rs` — `(kind, bytes)` became `Some((kind, bytes))` and the
  `supports_runtime_snapshot` override was deleted. No register, reset value,
  decode path or instruction timing is touched. Byte-for-byte the same snapshot
  blob and the same restore.
* `esp32c3/gpio.rs` — a `#[cfg(test)]` call site only, `.expect()` on the new
  `Option`.

This is the "source digest moved; the model did not" case the manifest already
documents for the 2026-08-13 acks. Clearing it is a one-line dated `drift_ack`
bump with a written reason, plus the regenerated `VALIDATION_STATUS.md`.

**I have not made that change.** A drift ack is a human claim about what silicon
was and was not re-verified, and moving one to make a gate go green is not mine
to do unreviewed — say the word and I will push the ack with the reasoning
above, or push a re-capture instead if you would rather discharge it properly.

Note the knock-on: a failed step skips the rest of its job, so the eight
`pr-gate` steps after this one (clippy, unit tests, AVR golden survival, browser
layer, firmware survival) did not run on this PR. They were run locally instead —
see the table in the description. `pr-workspace-tests` is a separate job and is
unaffected.

PR #995, opened 2026-08-15
fix(esp32s3): model the S3's interrupt matrix, not the C3's

w1ne:
## The defect

`configs/peripherals/esp32s3/interrupt_core0.yaml` did not describe the ESP32-S3's
interrupt matrix. Verified against the vendored SVD in this repo —
`tests/fixtures/svd/esp32s3.svd`, peripheral `INTERRUPT_CORE0`, base `0x600C_2000`
(byte-identical to `cmsis-svd-data/data/Espressif/esp32s3.svd` for this block):

| register | was | silicon |
|---|---|---|
| `PRO_MAC_INTR_MAP` | `0x000`, **named `MAC_INTR_MAP`** | `0x000` |
| `GPIO_INTERRUPT_PRO_MAP` | `0x040` | `0x040` |
| `SPI_INTR_2_MAP` | `0x04C` | `0x054` |
| `SPI_INTR_3_MAP` | `0x050` | `0x058` |
| `UART_INTR_MAP` | `0x058` | `0x06C` |
| `UART1_INTR_MAP` | `0x05C` | `0x070` |
| `UART2_INTR_MAP` | `0x060` | `0x074` |
| `LEDC_INT_MAP` | `0x068` | `0x08C` |
| `USB_INTR_MAP` | `0x088` | `0x098` |
| `TG_T0_INT_MAP` | `0x0A0` | `0x0C8` |
| `TG_T1_INT_MAP` | `0x0A4` | `0x0CC` |
| `TG_WDT_INT_MAP` | `0x0A8` | `0x0D0` |
| `TG1_T0_INT_MAP` | `0x0AC` | `0x0D4` |
| `TG1_T1_INT_MAP` | `0x0B0` | `0x0D8` |
| `TG1_WDT_INT_MAP` | `0x0B4` | `0x0DC` |
| `CLOCK_GATE` | `0x1A0` | `0x19C` |

One cause: the S3 has `GPIO_INTERRUPT_APP_MAP` (0x48), `GPIO_INTERRUPT_APP_NMI_MAP`
(0x4C) and `SPI_INTR_1_MAP` (0x50); the C3 has none of them. Those three rows were
missing, so every register below them sat 8, then 20, then 40 bytes low. They are
added here.

Three more, in the same file:

* **Name.** Source 0 was `MAC_INTR_MAP` — the C3's name. The S3 calls it
  `PRO_MAC_INTR_MAP`. Because the names differed, no tool ever compared the two
  maps: the mismatch hid itself.
* **Reset values.** Every `*_MAP` register resets to `0x10` on this part (SVD
  `resetValue`). The model said `0x0` for all of them.
* **Registers that do not exist on this silicon.** `0x104..0x19C` held
  `CPU_INT_ENABLE` / `CPU_INT_TYPE` / `CPU_INT_CLEAR` / `CPU_INT_EIP_STATUS` /
  `CPU_INT_PRI_0/1` / `CPU_INT_THRESH` and `INTR_STATUS_REG_0/1` — the C3's RISC-V
  INTC control block, at the C3's offsets. An Xtensa LX7 part has none of them, and
  on the S3 those addresses are ordinary source-map registers. Replaced with what
  the silicon puts there: `PRO_INTR_STATUS_0..3` at `0x18C..0x198` — which is also
  exactly where the behavioural model reads them (`INTR_STATUS_BASE` in
  `crates/core/src/peripherals/esp32s3/intmatrix.rs`).

The `interrupts:` block also carried the C3's source IDs for `FROM_CPU_INTR0..3`
(50..53). On the S3 `CPU_INTR_FROM_CPU_0_MAP` is at `0x13C`, so the sources are
79..82 — which is already what `peripherals/esp32s3/crosscore_ipi.rs` uses.

## Corrections to the report this started from

Re-derived from the SVD before changing anything. Three details in the original
write-up are wrong; none of them changes the fix, but they should not be repeated:

1. **`INTERRUPT_CORE0` is 105 registers, not 101, and is not fully contiguous.**
   `0x000..0x19C` is contiguous (104 registers), then `DATE` sits alone at `0x7FC`.
2. **`UART_INTR_MAP` on the C3 is at `0x54`, not `0x58`.** `0x58` is the C3's
   `UART1_INTR_MAP`.
3. **`TG_T0_INT_MAP` on the C3 is at `0x80`, not `0xA0`.** `0xA0` is the C3's
   `SPI_MEM_REJECT_INTR_MAP`.

So "every wrong offset is exactly the C3's offset" holds for `SPI_INTR_2_MAP`
(0x4C) and for the whole `CPU_INT_*` block, but not for the rest: the file was a
hand-compressed hybrid, not a copy of the C3 map. The defect itself — wrong
offsets, wrong first-register name, wrong reset values, C3-only registers — is
confirmed exactly as described.

## Blast radius: none at runtime

`configure_xtensa_esp32s3` calls `bus.peripherals.clear()`
(`crates/core/src/system/xtensa/esp32s3.rs:379`) before registering
`Esp32s3IntMatrix` at `0x600C_2000`. Every S3 firmware path — CLI `run`/`test`,
wasm, world nodes, DAP, Python — goes through that builder, so the declarative
descriptor is *deleted*, not out-voted, and the behavioural model (which already
decoded at `4 * source_id`) is what firmware talks to. This descriptor is live only
on `from_config`-built buses: `chip_conformance`, `bus_visibility`,
`unknown_peripheral_type`.

That makes this a correction to a committed model artifact, not a change in what
any simulation does today. It is still worth fixing: the file is the S3 interrupt
matrix's declarative model of record, it is `include_str!`d into every native and
wasm binary, it is watched by the board drift gate, and it is one `type:` change
away from being the live model.

**The ESP32-C3 is untouched** and stays 103/103 against its own SVD. Every
construction site was checked: the only cross-chip coupling is
`bus/routing.rs:331-334`, which binds the C3 INTC cache by name+base
(`p.name == "interrupt_core0" && p.base == 0x600C_2000`) — which the S3 chip YAML
also declares. That cache is only consumed when `esp32c3_irq_routing` is set, and
that happens in `boot/esp32c3_rom.rs` alone.

## Prove it was broken: RED before, GREEN after

New gate: `crates/core/tests/esp32s3_interrupt_map_vendor_offsets.rs`. Nothing
compared a peripheral descriptor's register offsets to an SVD before —
`svd_conformance` checks base addresses and IRQ numbers only, and
`register_coverage` counts *responses*, which a register at the wrong offset still
produces. An undecoded write in a declarative peripheral is a **silent no-op**, so
"nothing happens" is also what an untested fix looks like.

RED — the descriptor reverted to its committed content (`git diff` empty for that
file, i.e. provably back at HEAD; `grep` confirmed `UART_INTR_MAP: address_offset:
0x058`), everything else unchanged:

```
running 3 tests
test esp32s3_interrupt_matrix_decodes_at_the_vendor_offset ... FAILED
test esp32c3_interrupt_map_is_unchanged_at_its_own_offsets ... ok
test esp32s3_interrupt_map_matches_the_vendor_svd ... FAILED

---- esp32s3_interrupt_matrix_decodes_at_the_vendor_offset stdout ----
assertion `left == right` failed: UART_INTR_MAP (0x6C) must reset to 0x10; 0 means
the offset is undecoded (silent no-op) or the reset value is still the old 0
  left: 0
 right: 16

---- esp32s3_interrupt_map_matches_the_vendor_svd stdout ----
configs/peripherals/esp32s3/interrupt_core0.yaml disagrees with
tests/fixtures/svd/esp32s3.svd (45 findings of 32 registers checked):
  MAC_INTR_MAP: not a register of this silicon's INTERRUPT_CORE0 (offset 0x000 in our model)
  MAC_NMI_MAP: reset 0x00000000, silicon resets to 0x00000010
  SPI_INTR_2_MAP: modelled at 0x04C, silicon puts it at 0x054
  UART_INTR_MAP: modelled at 0x058, silicon puts it at 0x06C
  TG_T0_INT_MAP: modelled at 0x0A0, silicon puts it at 0x0C8
  CPU_INT_ENABLE: not a register of this silicon's INTERRUPT_CORE0 (offset 0x104 in our model)
  CPU_INT_THRESH: not a register of this silicon's INTERRUPT_CORE0 (offset 0x194 in our model)
  CLOCK_GATE: modelled at 0x1A0, silicon puts it at 0x19C
  … 37 more

test result: FAILED. 1 passed; 2 failed
```

GREEN — with the fix restored (`git diff --stat` non-empty: `1 file changed, 118
insertions(+), 93 deletions(-)`):

```
running 3 tests
test esp32s3_interrupt_matrix_decodes_at_the_vendor_offset ... ok
test esp32c3_interrupt_map_is_unchanged_at_its_own_offsets ... ok
test esp32s3_interrupt_map_matches_the_vendor_svd ... ok

test result: ok. 3 passed; 0 failed
```

And the C3 guard can say no. Sabotage: `esp32c3/interrupt_core0.yaml`
`UART_INTR_MAP` moved 84 → 108 (i.e. onto the S3's `0x6C`). `git diff` non-empty
before the run (`1 file changed, 1 insertion(+), 1 deletion(-)`, diff hunk shown):

```
test esp32s3_interrupt_matrix_decodes_at_the_vendor_offset ... ok
test esp32c3_interrupt_map_is_unchanged_at_its_own_offsets ... FAILED
test esp32s3_interrupt_map_matches_the_vendor_svd ... ok

configs/peripherals/esp32c3/interrupt_core0.yaml disagrees with
tests/fixtures/real_world/esp32c3.svd (1 findings of 103 registers checked):
  UART_INTR_MAP: modelled at 0x06C, silicon puts it at 0x054
```

Sabotage reverted; `git diff` for that file empty again.

## Regeneration

`scripts/regen-all.sh` run: exit 0, and the only artifact it moved is
`docs/boards/VALIDATION_STATUS.md`. A second run left the tree clean, so nothing
else is stale. `embedded_descriptors.rs` did **not** change — it `include_str!`s
this file by path, so a content edit needs no regeneration there.

`generate_validation_status.py` digests every path in the board's `models` list,
which includes `configs/peripherals/esp32s3`, so the esp32s3 row flipped to
`✖ DRIFT — model 2026-08-13 > capture; RE-CAPTURE` and the required PR-gate step
`generate_validation_status.py --check --drift` exited **1** (measured). The ack is
therefore re-stamped in `validation/manifest.yaml` with a dated reason: the
2026-08-09 capture read UART0/GPIO/I2C0/RMT/MCPWM0/TIMG0/SYSTIMER/GDMA/SYSTEM/
RTC_CNTL — `INTERRUPT_CORE0` is not among them, so no captured value moves and a
re-capture could not evidence this change either way. `--write-ack-digests` stamped
exactly 1 board. Gate now exits 0.

## Scope — what is deliberately NOT modelled

This models the source rows the S3's own models drive, 31 registers, not all 99.
Offsets not listed stay undecoded — read-as-zero, writes dropped — exactly as
before; that gap is pre-existing and unchanged by this PR. `I2C_EXT0_INTR_MAP`
(0xA8, source 42) is added because the Tier-1 S3 fixture binds that row through the
matrix (`examples/tier1-fixture/esp32s3/src/main.rs`, `check_irq`) — a row with a
proven consumer, not a row added to raise a count.

Left out on purpose:

* `DATE` (0x7FC) — an informational silicon-revision stamp outside the contiguous
  block; the behavioural model reads-as-zero above 0x19C.
* The `INTERRUPT_CORE1` mirror at +0x800 — the SVD carries it as its own
  peripheral, and `Esp32s3IntMatrix` already serves that window.
* The 68 source rows between the modelled ones. Modelling all 99 is the obvious
  follow-up; it is a bigger claim than this fix and belongs in its own change.

## Verification (measured, by exit code)

| suite | pristine `main` | this branch |
|---|---|---|
| `cargo test -p labwired-core --lib` | 3093 passed, 0 failed, exit 0 | 3093 passed, 0 failed, exit 0 |
| `chip_conformance`, `bus_visibility`, `unknown_peripheral_type`, `register_coverage`, `svd_conformance` | 13 passed, 0 failed, exit 0 | 13 passed, 0 failed, exit 0 |
| new `esp32s3_interrupt_map_vendor_offsets` | — | 3 passed, 0 failed, exit 0 |
| `intmatrix_alarm`, `debug_schema_is_not_fidelity` | — | 2 passed, 0 failed, exit 0 |
| `esp32s3_walk_differential` + `intmatrix_alarm` (`--features event-scheduler`) — the S3 interrupt path on the behavioural model | — | 3 passed, 0 failed, exit 0 |

`cargo fmt --all -- --check` exit 0. `cargo clippy -p labwired-core --test
esp32s3_interrupt_map_vendor_offsets -- -D warnings` clean.

Not run here, for an environment reason rather than a result:
`e2e_hello_world` (`--features esp32s3-fixtures`) cross-compiles the S3 example
and dies at `error: linker xtensa-esp32s3-elf-gcc not found` — the Xtensa ESP
toolchain is not installed on this machine. That is not a verdict on the change;
CI has the toolchain.

### DIFF
--- a/configs/peripherals/esp32s3/interrupt_core0.yaml
+++ b/configs/peripherals/esp32s3/interrupt_core0.yaml
@@ -1,219 +1,244 @@
-# ESP32-S3 Interrupt Matrix - CPU Core 0
-# Based on ESP32-S3 Technical Reference Manual
+# ESP32-S3 Interrupt Matrix - CPU Core 0 (INTERRUPT_CORE0, base 0x600C_2000)
+#
+# Offsets, register names and reset values are taken from the vendored Espressif
+# SVD, tests/fixtures/svd/esp32s3.svd, peripheral INTERRUPT_CORE0. That map is a
+# uniform array: INTERRUPT_CORE0_<source>_MAP_REG sits at 4 * source_id, so an
+# offset here IS the source ID times four, and PRO_INTR_STATUS_0..3 / CLOCK_GATE
+# follow the last map register.
+#
+# ⚠️ This file previously carried the ESP32-C3's map, not the S3's. The C3 lacks
+# GPIO_INTERRUPT_APP_MAP (0x48), GPIO_INTERRUPT_APP_NMI_MAP (0x4C) and
+# SPI_INTR_1_MAP (0x50); with those three rows missing every later register was
+# modelled 8, then 20, then 40 bytes low (SPI_INTR_2_MAP 0x4C vs 0x54,
+# UART_INTR_MAP 0x58 vs 0x6C, TG_T0_INT_MAP 0xA0 vs 0xC8). The first register
+# was also named MAC_INTR_MAP — the C3's name; the S3 calls it PRO_MAC_INTR_MAP,
+# and because the names differed nothing ever compared the two maps. The block
+# at 0x104..0x194 held CPU_INT_ENABLE / CPU_INT_TYPE / CPU_INT_CLEAR /
+# CPU_INT_EIP_STATUS / CPU_INT_PRI_n / CPU_INT_THRESH: those are the C3's RISC-V
+# INTC control registers. The S3 is Xtensa LX7 and has none of them — those
+# offsets are ordinary source-map registers on this silicon.
+#
+# Coverage: this descriptor models the source rows the S3 models actually drive,
+# not all 99. Unlisted source offsets stay undecoded (read-as-zero, writes
+# dropped) exactly as before — that gap is pre-existing and unchanged here. The
+# behavioural S3 model is crates/core/src/peripherals/esp32s3/intmatrix.rs,
+# which decodes all 99 sources plus the CORE1 window at +0x800; this descriptor
+# must not disagree with it. Deliberately unmodelled: DATE (0x7FC, an
+# informational silicon-revision stamp) and the INTERRUPT_CORE1 mirror
+# (+0x800), which the SVD carries as its own peripheral.
+#
+# Every *_MAP register resets to 0x10 on this part (SVD resetValue), not 0x0.
 
 peripheral: "INTERRUPT_CORE0"
-version: "0.1.0"
+version: "0.2.0"
 
 registers:
-  - id: "MAC_INTR_MAP"
+  - id: "PRO_MAC_INTR_MAP"
     address_offset: 0x000
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
   - id: "MAC_NMI_MAP"
     address_offset: 0x004
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
   - id: "PWR_INTR_MAP"
     address_offset: 0x008
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
   - id: "BB_INT_MAP"
     address_offset: 0x00C
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
   - id: "BT_MAC_INT_MAP"
     address_offset: 0x010
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
   - id: "BT_BB_INT_MAP"
     address_offset: 0x014
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
   - id: "BT_BB_NMI_MAP"
     address_offset: 0x018
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
   - id: "GPIO_INTERRUPT_PRO_MAP"
     address_offset: 0x040
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
   - id: "GPIO_INTERRUPT_PRO_NMI_MAP"
     address_offset: 0x044
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
-  - id: "SPI_INTR_2_MAP"
+  # 0x048/0x04C/0x050 exist on the S3 and not on the C3. Their absence is what
+  # pushed every register below them onto a C3 offset.
+  - id: "GPIO_INTERRUPT_APP_MAP"
+    address_offset: 0x048
+    size: 32
+    access: "R/W"
+    reset_value: 0x00000010
+
+  - id: "GPIO_INTERRUPT_APP_NMI_MAP"
     address_offset: 0x04C
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
-  - id: "SPI_INTR_3_MAP"
+  - id: "SPI_INTR_1_MAP"
     address_offset: 0x050
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
-  - id: "UART_INTR_MAP"
+  - id: "SPI_INTR_2_MAP"
+    address_offset: 0x054
+    size: 32
+    access: "R/W"
+    reset_value: 0x00000010
+
+  - id: "SPI_INTR_3_MAP"
     address_offset: 0x058
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
+
+  - id: "UART_INTR_MAP"
+    address_offset: 0x06C
+    size: 32
+    access: "R/W"
+    reset_value: 0x00000010
 
   - id: "UART1_INTR_MAP"
-    address_offset: 0x05C
+    address_offset: 0x070
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
   - id: "UART2_INTR_MAP"
-    address_offset: 0x060
+    address_offset: 0x074
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
   - id: "LEDC_INT_MAP"
-    address_offset: 0x068
+    address_offset: 0x08C
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
   - id: "USB_INTR_MAP"
-    address_offset: 0x088
+    address_offset: 0x098
     size: 32
     access: "R/W"
-    reset_value: 0x00000000
+    reset_value: 0x00000010
 
-  - id: "TG_T0_INT_MAP"
-    address_offset: 0x0A0
-    size: 32
-    access: "R/W"
-    reset_value: 0x00000000
-
-  - id: "TG_T1_INT_MAP"
-    address_offset: 0x0A4
-    size: 32
-    access: "R/W"
-    reset_value: 0x00000000
-
-  - id: "TG_WDT_INT_MAP"
+  # Source 42. The Tier-1 S3 fixture binds this row through the matrix
+  # (examples/tier1-fixture/esp32s3/src/main.rs, check_irq), so it is a proven
+  # consumer of a decoded offset, not a row added to raise a count.
+  - id: "I2C_EXT0_INTR_MAP"
     address_offset: 0x0A8
     size: 32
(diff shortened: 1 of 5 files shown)


w1ne:
Cross-checked this against the **ESP32-S3 TRM v1.8** while adjudicating the 34 conflicts `0036_part_knowledge_svd_promotion.sql` published (labwired-core#1003 closes the TIMG0/GPIO nine). Two things worth having on the record here.

## 1. The TRM independently confirms every offset in this PR

This PR derived the map from the vendored SVD. The reference manual, which outranks an SVD, agrees on all of it — §9.4.1 CPU0 Interrupt Register Summary (pp.547–550) and §9.5.1 (pp.555–556):

| register | this PR | TRM v1.8 |
|---|---|---|
| `SPI_INTR_1_MAP` | 0x050 | 0x0050 ✓ |
| `SPI_INTR_2_MAP` | 0x054 | 0x0054 ✓ |
| `SPI_INTR_3_MAP` | 0x058 | 0x0058 ✓ |
| `UART_INTR_MAP` | 0x06C | 0x006C ✓ |
| `UART1_INTR_MAP` | 0x070 | 0x0070 ✓ |
| `UART2_INTR_MAP` | 0x074 | 0x0074 ✓ |
| `LEDC_INT_MAP` | 0x08C | 0x008C ✓ |
| `USB_INTR_MAP` | 0x098 | 0x0098 ✓ |
| `TG_T0_INT_MAP` | 0x0C8 | 0x00C8 ✓ |
| `TG_T1_INT_MAP` | 0x0CC | 0x00CC ✓ |
| `TG_WDT_INT_MAP` | 0x0D0 | 0x00D0 ✓ |
| `TG1_T0_INT_MAP` | 0x0D4 | 0x00D4 ✓ |
| `TG1_T1_INT_MAP` | 0x0D8 | 0x00D8 ✓ |
| `TG1_WDT_INT_MAP` | 0x0DC | 0x00DC ✓ |
| `INTR_STATUS_0..3` | 0x18C..0x198 | 0x018C..0x0198 ✓ |
| `CLOCK_GATE` | 0x19C | 0x019C ✓ |

The `0x10` reset is confirmed too, and by a source that is not the SVD: TRM p.557, Register 9.84, gives the shared figure for these registers as `INTERRUPT_CORE0_SOURCE_Y_MAP` (b4:0) **Reset 16**, with b31:5 reserved. So the eight `*_MAP` reset-value conflicts are the same root cause as the fourteen offset ones, verified rather than assumed. That was the specific thing I was asked to check rather than take on trust.

The TRM also confirms the diagnosis: its map jumps 0x2C → 0x38, 0x44 → 0x50, 0x80 → 0x8C and 0xAC → 0xC8. Those four gaps are exactly what a hand-compressed file loses.

## 2. One name in this PR follows the SVD against the TRM

TRM v1.8 §9.4.1 calls source 0 **`INTERRUPT_CORE0_MAC_INTR_MAP_REG`** — the name this PR is replacing — and calls 0x40/0x44 **`GPIO_INTERRUPT_CPU_MAP`** / **`GPIO_INTERRUPT_CPU_NMI_MAP`**, not `..._PRO_...`. The SVD says `PRO_MAC_INTR_MAP` and `GPIO_INTERRUPT_PRO_MAP`.

Not asking for a change: taking the SVD's spelling is what makes these registers *comparable* at all — under the TRM name, source 0 is invisible to the promotion pass and its reset value goes unchecked, which is precisely how the original mismatch hid itself. But "the S3 calls it `PRO_MAC_INTR_MAP`" in the PR body is the SVD's claim, not the reference manual's, and the two disagree. Worth saying so in the file rather than leaving a future reader to rediscover it.

Also note the TRM does **not** list `GPIO_INTERRUPT_APP_MAP` (0x48) or `GPIO_INTERRUPT_APP_NMI_MAP` (0x4C) at all — its summary skips 0x48/0x4C. They are in the SVD, and adding them is what restores the alignment, so the PR is right to add them; they are simply undocumented in TRM v1.8.

## 3. Merge-order note

labwired-core#1003 also re-stamps the esp32s3 `drift_ack` in `validation/manifest.yaml`, so these two PRs will conflict there textually. Whichever lands second needs its `drift_ack_digest` re-stamped (`python3 scripts/generate_validation_status.py --write-ack-digests`) because the digest covers `configs/peripherals/esp32s3`, which both PRs move.

Measured end state with both applied: `npx tsx packages/api/scripts/generate-svd-promotion.ts --report` reports **esp32s3: 39 promoted to proven, 0 published as conflicts** (from 3 proven / 31 conflicts on main). All 31 of the ESP32-S3 conflicts close; the remaining 3 corpus-wide are RP2040 `SYSINFO` ×2 and MKW41Z4 `SIM.SDID`, and neither is a model defect.

PR #1008, opened 2026-08-17
ESP32-S3: octal PSRAM, dual-core startup, and LCD_CAM i80 pixel streaming

w1ne:
Raises the ESP32-S3 twin far enough to boot and run a real, unmodified ESP-IDF
application: dual core, 16 MB flash, 8 MB octal PSRAM, a memory-mapped data
partition, and a parallel LCD driven through LCD_CAM + GDMA.

The test subject is Doom — doomgeneric with the shareware IWAD in its own raw
flash partition — chosen because it exercises all of that at once and because
the same firmware runs on physical hardware, so silicon is available as an
oracle throughout.

## What was broken

Each of these independently stopped a real firmware, and none were specific to
the test subject:

| defect | who it broke |
|---|---|
| No PSRAM device on MSPI CS1 | every `CONFIG_SPIRAM=y` S3 image — `esp_psram_init` aborted at `cpu_start` |
| JEDEC capacity hardcoded to 4 MB | every 8/16 MB image — `esp_flash` refused to boot |
| `esp32s3.yaml` said `16MB` | `parse_size` treats KB as binary and MB as decimal → an 8 MiB part |
| `WAITI` did not advance the PC | **every dual-core ESP-IDF image** — core 1 woke and returned into the wait forever |
| `LC_DMA_INT_*` at `0x34..0x40` | **every i80 panel bring-up** — silicon has it at `0x64..0x70` |
| LCD_CAM interrupt bits at 0/2/4/6 | silicon packs them 0/1/2/3 |
| No LCD_CAM pixel path | parallel panels could not paint |
| APP_CPU release lived in one frontend | core 1 was constructed but never released outside the native CLI |

The `WAITI` one is the deepest. Core 1 was **not** starved of interrupts — it
woke 570 times in 60 M steps, on the 1 kHz systimer tick, and ran real
handlers. Because the PC never advanced, `EPC1` pointed at the `WAITI` itself
and every return re-entered the wait. The Xtensa ISA says `WAITI` retires and
then waits.

## Evidence

Booting a real ESP-IDF v5.3.1 image from a merged flash image through the
faithful mask ROM:

- PSRAM probe output matches the physical board line for line, including the
  MSPI timing tuning index and `Adding pool of 7957K`, and `SPI SRAM memory
  test OK` proves IDF wrote and read external RAM through the MMU.
- `internal free 314519 B, PSRAM free 8146120 B` and `IWAD mapped at
  0x3c880000: 1264 lumps, 4196020 bytes` are byte-identical to the board's own
  serial output.
- The panel paints Doom: `ink_bytes=123216/153600`, with region *and* ink
  bounded — the two 20-row letterbox bands are exactly 0 lit and the 320x200
  image area is 98.9% lit, so nothing bled outside the CASET/PASET window.
- Panel content matches the thumbnail the firmware itself prints on serial for
  the same frame: 1529/1536 characters exact, 100% within one ramp step (the
  residue is RGB565 vs ARGB8888 quantisation).

Every fix carries a negative control. Removing the one-line `WAITI` change
reproduces the original hang at the identical PC; restoring `INT_ENA` to
`0x34` fails 4 of the 5 new pixel tests; stubbing `dma_push_tx` fails 3 of 5.

## Tests

`cargo test -p labwired-core --no-fail-fast`: **4014 → 4021 passing**.

Three targets are red **on `main` already** and are untouched here:
`cpu_trace_conformance::every_cpu_core_is_covered_by_this_file` (an `Avr` core
exists that the test's `cores()` list does not name),
`e2e_nrf52840_obd2_scanner` (scanner ELF not built in this environment) and
`strict_onboarding` (stm32l073 firmware build failure, no atmega328p example
dir). Plain `cargo test` stops at the first failing target and reports only
~18 of 141, so `--no-fail-fast` is the honest comparison.

New: `crates/core/tests/esp32s3_lcd_i80_pixels.rs` (5 tests) drives the whole
path through `SystemBus` at silicon addresses — no model is called directly —
and `crates/core/src/tests/xtensa_waiti_resume.rs` (2 tests) asserts `EPC1` and
that the instruction after a `WAITI` actually executes.

## CLI

`labwired run` gains `--system` (the S3 rom-boot path builds its bus in Rust
and never ran `from_config`, so it had no way to attach manifest devices),
`--display-out` and `--stop-on` (frame-exact evidence instead of a hand-tuned
`--max-steps`).

## Known gaps, deliberately not addressed here

- **The browser stalls on a dual-core image.** `crates/wasm` builds core with
  `event-scheduler`; `crates/cli` deliberately does not, so the two engines
  take different idle paths. Under wasm the guest spins after `log_hardware`
  and burns 2e9 cycles without progress, where native needs ~66 M to pass the
  same point. A unicore image gets much further in the same engine, so the
  differentiator is dual core under that feature. Reproducible natively with
  `--features event-scheduler`. Under investigation; this PR does not claim
  browser support.
- `bus/policy.rs::max_safe_tick_interval` allows 512 whenever the legacy walk
  is disabled; on the S3 with an ESP-IDF image that is false — 512 livelocks in
  `xTaskIncrementTick` and 16 crashes. The browser applies 512 today.
- No LCD_CAM pixel clock or line timing (a transaction moves the whole chain
  then completes), no RGB-mode sync generation, no camera DVP RX path.
- The ESP32-classic LX6 path still pre-paints its own dual-core handshake
  flags. Out of scope; the `WAITI` fix may make it unnecessary there too.
- `Ili9341::oriented_framebuffer` (the SPI twin) has the same `MADCTL_MV`
  shear fixed here on the parallel twin.


w1ne:
**Update: the browser runs Doom.** The stall this PR documented as a known gap is fixed and included here.

It was neither of the two things the description guessed at. Not `event-scheduler` — a CLI built with that feature reaches frame 1 and paints. Not `esp_partition_mmap` — the fixed run maps the partition fine. The guest was never hung either; it made ~2.1 MIPS of forward motion the whole time.

**Root cause.** ESP-IDF implements `portYIELD_WITHIN_API()` as `esp_crosscore_int_send_yield(xPortGetCoreID())` — ring your *own* core's doorbell inside a critical section and expect preemption the moment `portEXIT_CRITICAL` re-enables interrupts. Deliver that late and `xQueueReceive`'s `for(;;)` comes round again and calls `vTaskPlaceOnEventList` a second time for a task already on that list; `vListInsert` links the item to itself and the next insert walks the loop forever, while the other core spins in `spinlock_acquire` behind a queue lock that is never released.

Four engine defects delivered it late:

1. The ESP32-S3 had **no MMIO write choke** — the routed per-core IRQ mask was rebuilt only by per-tick aggregation. The C3 has had this since its walk-free migration; the S3 twin was simply never written.
2. `plan_cpu_window`'s `SECONDARY_PARKED` clause ran the primary up to 1024 instructions per boundary *even at `peripheral_tick_interval == 1`* — which its own comment advertised.
3. A dual-core machine was allowed to relax the tick interval at all. The bus predicate bounds peripheral *observation* error; on SMP, interrupt-delivery skew is a *semantic* error.
4. The wasm S3 flash-boot constructor set `faithful_windows` on the PRO core only, so core 1 restored a garbage SP and faulted on every boundary — invisible, because `Sim::step_batch` returns `Ok(elapsed)` whenever the primary retired a cycle.

The differentiator between native and browser was never the feature flag: native `labwired run` drives `AdvanceRequest::single()`, the browser drives `AdvanceRequest::run(..)`.

**Verified in the browser engine**, dual core, parallel panel, real 4.2 MB WAD, faithful ROM boot:

```
I (213) cpu_start: Multicore app
I (271) doom: ESP32-S3 rev v0.0, 2 core(s), 16 MB flash
I (275) doom: IWAD mapped at 0x3c880000: 1264 lumps, 4196020 bytes
I (787) doom_video: panel up: 320x240 i80 16-bit at 10000000 Hz
I (1011) doom_video: frame 1  hash 0xec236c72  ink 100.0%
I (1031) doom: doom initialised, entering tick loop
```

That frame hash `0xec236c72` is the same one produced by the native simulator, by the physical ESP32-S3-N16R8 running this ESP-IDF firmware, and by an independent Arduino-framework port of the same game on the same board. Same pixels in four places.

**Also fixed here, same root cause:** the tick-interval defect. `Machine::max_safe_tick_interval` clamps to 1 only when a secondary CPU is present, and the browser now asks the machine rather than the bus. Single-core boards — unicore S3, C3, nRF, RP2040, STM32 — are untouched and keep `RECOMMENDED_TICK_INTERVAL`. Dual-core S3 labs run at ~0.5–2.4 MIPS where they previously did not run at all.

**And separately:** `Ili9341::oriented_framebuffer` (the SPI twin) had the same `MADCTL_MV` shear fixed earlier on the parallel twin — landscape pixel (300,0) was dropped entirely, (0,1) landed 80 px into row 0. Portrait output is byte-identical.

**New gate:** `crates/core/tests/esp32s3_smp_yield_latency.rs`, registered in the merge-gating `pr-scheduler-observable` lane, with negative controls verified red for each of the three engine changes.

**Correction to the test baseline in the description:** there are more pre-existing red targets than it claims — four on default features (the fourth is `--lib`'s `downcast_ratchet`, "as_any() call sites rose to 194, ceiling 193", identical on both sides) and seven under `--features event-scheduler`. Identical on both sides of this diff in both configurations.

**Worth its own issue, not fixed here:** `Sim::step_batch` masks CPU errors whenever the primary retired at least one cycle. A secondary faulting on *every* boundary read as steady progress in the browser, which is why this looked like a hang for hours.

PR #1010, opened 2026-08-17
fix(validation): the board digest could not see the CPU it validates

w1ne:
Closes ledger row **F-3**. Found while unblocking core#1007's `pr-gate`.

## The gate was blind to the core

`docs/boards/VALIDATION_STATUS.md` exists to answer *"is the model right vs silicon?"*, and its models column is *"a content digest over everything that board's `models` list names."*

Of 23 boards, those lists named a CPU **exactly twice** — `cpu/riscv.rs` on esp32c3, `cpu/xtensa_jit` on esp32s3 — and named `cpu/cortex_m.rs` or `decoder/arm.rs` **zero** times, against **109** peripheral paths.

Measured directly: append one line to `crates/core/src/cpu/cortex_m.rs`, regenerate, count digests that move.

| manifest | digests that noticed |
| --- | ---: |
| as on `origin/main` | **0** |
| with this fix | **19** |

A change to Thumb decoding for the entire ARM fleet was invisible to the drift gate.

I only found it because core#1007 — which changes exactly that — *also happened* to add two constants to `peripherals/scb.rs`, which `nrf5340` alone happens to list. Without that coincidence the gate would have reported no drift.

## Every board now names its own core and decoder

| arch | added | boards |
| --- | --- | ---: |
| arm | `cpu/cortex_m.rs`, `decoder/arm.rs` | 19 |
| riscv | `cpu/riscv.rs`, `decoder/riscv.rs` | 2 |
| xtensa | `cpu/xtensa_lx7.rs`, `decoder/xtensa.rs` | 2 |

**All 23 were missing something**, not only the ARM ones: esp32c3 carried its CPU but not its decoder; esp32s3 carried the JIT but neither the interpreter nor the decoder.

Xtensa maps to `xtensa_lx7.rs` for both esp32 and esp32s3 — the classic LX6 reuses it, as `system/xtensa/esp32.rs:102` says in as many words. Verified rather than assumed: both `configure_xtensa_esp32` and the S3 system construct `XtensaLx7`.

The arch is read from each board's own `configs/chips/<chip>.yaml` `arch:` field rather than a hand-written list, and **every path is asserted to exist before being written**, so a rename fails the generator instead of silently digesting nothing.

## Why all 23 digests move here

The covered set genuinely grew, so every digest changes once, in this commit. That is the intended one-time cost of the fix, not drift.

- `generate_validation_status.py --check` — clean
- `tests/board_registry_consistency` (the manifest's only other consumer) — 3/3

## Not in this PR

This makes the digest *notice* a CPU change. It does not add silicon-diff evidence for any board — the tier column is untouched, and a board that was `⚪ structural` still is.

### DIFF
--- a/docs/boards/VALIDATION_STATUS.md
+++ b/docs/boards/VALIDATION_STATUS.md
@@ -9,29 +9,29 @@ The models column is a content digest over everything that board's `models` list
 
 | Board | Tier | Last silicon capture | Models | Status |
 |-------|------|----------------------|--------|--------|
-| `nrf52840` | 🟢 silicon-verified | 2026-08-09 | `7ff611a215be6e82` | ⚠ drift acked 2026-08-13 (re-capture pending) |
-| `seeed-xiao-nrf52840-sense` | 🟢 silicon-verified | 2026-08-09 | `7ff611a215be6e82` | ⚠ drift acked 2026-08-13 (re-capture pending) |
-| `stm32h563` | 🟢 silicon-verified | 2026-08-10 | `37b51e405aa21480` | ⚠ drift acked 2026-08-13 (re-capture pending) |
-| `esp32c3` | 🟢 silicon-verified | 2026-08-09 | `235cfa41817e6ee6` | ⚠ drift acked 2026-08-15 (re-capture pending) |
-| `nucleo-l476rg` | 🟢 silicon-verified | 2026-08-09 | `d426d9ffbafaa978` | ⚠ drift acked 2026-08-13 (re-capture pending) |
-| `nucleo-l073rz` | 🟢 silicon-verified | 2026-08-09 | `911aa0ddc97db339` | ⚠ drift acked 2026-08-13 (re-capture pending) |
-| `stm32f103` | 🟢 silicon-verified | 2026-08-09 | `7f49451fc56984bf` | ⚠ drift acked 2026-08-13 (re-capture pending) |
-| `stm32f407` | 🟢 silicon-smoke | 2026-06-20 | `20b0f7311d064656` | ⚠ drift acked 2026-08-13 (re-capture pending) |
-| `esp32s3` | 🟢 silicon-verified | 2026-08-09 | `00b4461832f677ee` | ⚠ drift acked 2026-08-17 (re-capture pending) |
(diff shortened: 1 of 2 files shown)


w1ne:
## ⚠️ `pr-gate` is red, and it is right — this needs a human decision, not a patch

```
ERROR: silicon validation has DRIFTED (model changed after last capture, no covering drift_ack)
  nrf52840, seeed-xiao-nrf52840-sense, stm32h563, esp32c3,
  nucleo-l476rg, nucleo-l073rz, stm32f103, stm32f407, esp32s3
```

Nine boards with a real silicon capture now report drift. **Nothing about the models changed** — this PR only widens what each board's digest *covers*. But the drift rule compares the newest commit date across `models` to `silicon.last_capture`, and the newly-covered `cpu/cortex_m.rs` and `decoder/arm.rs` have been edited since those captures were taken.

**That is a true statement that was previously invisible: these nine boards' silicon captures do not cover their CPU model, and never did.** The capture validated peripherals against real hardware while the thing executing every instruction sat outside the digest. Widening the coverage is what surfaced it.

### Why I have not silenced it

`write_ack_digests` says so in its own docstring:

> Writing a digest is not an acknowledgement — the ack is the human assertion, the digest only pins WHICH tree it was asserted about. A board with no ack must stay un-acked; this must never be a way to bulk-silence the gate.

Stamping nine `drift_ack`s would be me asserting, on someone else's behalf, that an un-re-validated CPU model is accepted against silicon. That is not mine to assert.

### The three ways forward

| | what it means | cost |
| --- | --- | --- |
| **A. Ack the nine** | dated `drift_ack` + reason: "coverage widened to include the CPU; not re-validated against silicon, accepted pending re-capture" | honest and cheap; records a real gap instead of hiding it |
| **B. Re-capture** | run the live diff on bench hardware for all nine | real validation; needs the boards |
| **C. Narrow this PR** | add the CPU only to boards with no silicon capture | keeps the gate green and puts the gap back out of sight — I do not recommend it |

**A** is my recommendation, but the ack text is a claim about hardware validation and should carry your name, not mine.

The rest of the PR stands on its own: `--check` clean, `board_registry_consistency` 3/3, and the 0→19 measurement is unaffected.

PR #1036, opened 2026-08-20
test(ci): 18 test binaries are executed by NO workflow — gate the gap they sit in

w1ne:
`no_vacuous_test_targets.rs` states the contract plainly: **no test target may report success while executing zero tests.** It catches the `#![cfg(...)]` route, where the file compiles away. It does **not** catch the `#[ignore]` route, which reaches the identical end state — `0 passed`, exit 0, green tick — through a different door.

**Twenty targets sit in that gap. Eighteen are run by no workflow at all**, with or without `--ignored`. They compile into every PR shard and execute nothing, anywhere, ever.

```
esp32c3_clamped_full_state_differential   esp32c3_shipped_lab_batch_gate
tier1_matrix        tier1_matrix_ratchet  jit_lockstep
e2e_esp32c3_snapshot_resume               foreign_firmware_probe
```

A ratchet that never runs is not a ratchet.

## Shape

`runs_nothing_in_pr` in `workspace-test-shards.json` — a **shrink-only baseline, not permission**. Each entry carries `covered_by`, and eighteen say `NOTHING`: the finding stated in the data rather than in a comment someone can skip.

The aggregate now fails **both ways**:
- a target that runs nothing and is **not** baselined → the gap cannot reopen;
- a baselined target that **has** started running → the list cannot rot into an excuse for tests fixed long ago.

## Derived from execution, not from a regex

A source scan over `#[test]`/`#[ignore]` reported **41** candidates against the reports' **20** — `#[ignore]` also lands on non-test items, and required-features targets never build in a shard. Execution data is the only honest input.

## Verification

| | |
|---|---|
| the three real shard reports | 20 baselined, 18 dead, **0 problems** |
| negative control 1 — inject a new zero-executing target | **FAILS**: `RUNS NOTHING` |
| negative control 2 — flip a baselined target to passing | **FAILS**: `STALE BASELINE` |

Deletes nothing. Each of the eighteen is now a listed bug to close by running it or removing it.

### DIFF
--- a/scripts/ci/workspace-test-shards.json
+++ b/scripts/ci/workspace-test-shards.json
@@ -2,76 +2,259 @@
   "schema": 1,
   "_about": "Classification data for scripts/ci/workspace_test_shard.py and workspace_test_aggregate.py, driven by the pr-workspace-tests jobs in .github/workflows/core-ci.yml. Every default-runnable test target in the workspace runs in one of the PR shards EXCEPT the cross_build_excluded entries below. A target not covered by either list is a hard error, so a new test file cannot silently miss the PR lane.",
   "shard_count": 3,
-
   "_cross_build_about": "Targets that cannot pass on a runner with no cross toolchain: they build firmware for thumbv*/xtensa at test time, or they panic when a pre-built cross ELF is missing. core-full (nightly / [full-ci] / dispatch) installs the five ARM targets, builds the fixtures, and runs these via plain `cargo test --workspace`, so excluding them from PR shards removes NO coverage — it moves it to the lane that can actually execute it. A graceful-skip entry is excluded anyway: a test that skips on every single PR provides no signal, and keeping it would train the skip channel to carry noise.",
   "cross_build_excluded": [
-    { "package": "labwired-core", "target": "e2e_stm32f407_i2c",
+    {
+      "package": "labwired-core",
+      "target": "e2e_stm32f407_i2c",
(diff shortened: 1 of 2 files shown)


w1ne:
Closing this. The finding behind it is wrong, and wrong in the direction that would have cost coverage.

**`tier1_matrix` and `tier1_matrix_ratchet` are not dead.** They carry `#[cfg_attr(debug_assertions, ignore = "tier1 matrix sims run in release (see core-ci.yml tier1 step)")]` and are run by `core-ci.yml:1514` in the release lane. My proximity grep for `--ignored` missed that command because it does not need the flag — in release the ignore does not apply. This PR would have baselined a working, gating test as "executed by NO workflow".

**The rest are already registered, better than this PR does it.** `docs/testing/IGNORED_TESTS.md` is GENERATED by `scripts/generate_ignored_tests.py`, tracks all 114 `#[ignore]` attributes across 59 files, records which carry a reason (86 of 114), and has its own shrink-only ratchet at 28 undocumented with `--check` failing on staleness. The ones I flagged carry explicit reasons like `"diagnostic; run with --release --ignored --nocapture"`.

And the aggregate already reports a `release-only targets (empty in debug by construction)` category, which is exactly the population this PR re-counted under a scarier name.

So the gap I described is mostly not a gap, the part that is real is already ratcheted somewhere better, and shipping this would have added a second register that disagreed with the first. Nothing to fix here.

PR #1040, opened 2026-08-21
feat(efr32mg26): verify the model against the die over SWD

w1ne:
BRD2709A had **no silicon capture at all**. Every register in this model came from the vendor CMSIS headers — authoritative for the *map*, not the same as measured silicon. It is now the **second chip in `chip_conformance` with a real capture**, after esp32c3, and moves **L0 → L1 at 178/216 (82%)**.

## Harness

`hw_conform.sh` gains an `openocd-jlink` transport — the SEGGER J-Link OB every Silicon Labs kit carries. No new script; the existing standard already produces `reg_oracle.json` and the ratchet already diffs against it.

⚠️ **probe-rs is deliberately not used**, even though the rest of this board's tooling is. Its CLI has no reset-halt, and `--connect-under-reset` times out on BRD2709A — the probe's nRESET is not wired through. Without a halt at reset you capture whatever firmware left behind:

```
beacon running     CLKEN0/1/2 = c4010010 16014bff 00000080
openocd reset halt CLKEN0/1/2 = 00000000 00000000 00000000
```

⚠️ **The target carries a preamble, and that is silicon, not a shortcut.** A Series-2 peripheral that is not clocked does not read zero over the debug port — it **faults**, and openocd abandons the rest of its command list. A bare `reset halt` returns the CMU window and dies at GPIO. So the target un-gates `CLKEN0` and nothing else, the oracle records `state: reset_halt+preamble` instead of claiming a pure reset, and **`chip_conformance` replays those same writes into the sim**. Without the replay you compare an un-gated die against a gated model, every clock-gated peripheral reads as a "gap", and the real gaps drown in it.

## What the die corrected

**CMU — where this file had written down that it was guessing.** The module doc said *"LOCK read-back is the stored key … there is no bench capture for it."* There is now:

```
at reset               STATUS=0x00000001  LOCK=0  WDOGLOCK=0
after writing 0xdead   STATUS=0x80000001  LOCK=0
after the unlock key   STATUS=0x00000001  LOCK=0
```

- `LOCK`/`WDOGLOCK` are **write-only** — the key is never readable; lock state lives in `STATUS` bits 31/30.
- **The watchdog clock config boots UNLOCKED.** This file inferred "locked" from `_CMU_WDOGLOCK_RESETVALUE` being `0x5257` rather than the unlock key — but that's the LOCKKEY *field's* reset, not a lock state, and STATUS bit 30 is clear on the die.
- `STATUS` reads `0x00000001` at reset. `_CMU_STATUS_RESETVALUE` says `0x00000000` — **the die and the header disagree, and the die wins.**

**I2C — header and die both against the model:**

| register | model was | header + die |
|---|---|---|
| `STATE` | 0 | `0x01` (BUSY) |
| `STATUS` | `0xC0` (TXBL\|TXC) | `0x80` (TXBL alone) |
| `IF` | `0x18` | `0` |

`STATE` resetting to BUSY reads like a quirk and isn't: a controller nothing has driven hasn't learned where the bus is — which is exactly why emlib's `I2C_Init` opens with an ABORT. Modelled as `bus_idle_known`, cleared at reset and set by STOP/ABORT. `TXC` means a transfer completed, and out of reset none has.

```
cargo test -p labwired-core --lib             3228 passed
cargo clippy --all-targets -- -D warnings     exit 0
generate_validation_status.py --check --drift exit 0
```

Drift re-acked with the reason: the `i2c.rs` changes are all inside `Efr32s2I2c`, a variant the STM32 boards never construct — only the file's digest moved.

## The remaining tail, enumerated

26 registers still differ, split by whether the silicon value is a **documented** reset value in the vendor header:

- **Documented → real model gaps:** IADC `CFG`/`SCALE`/`FIFOCFG` and the `0x30..0x90` block, TIMER0/1 `IPVERSION`/`TOP`/`CNT`, USART0 `IPVERSION`/`FRAME`/`STATUS`/`IF`, I2C0 `IPVERSION`.
- **Not documented → per-die trim or warm boot state, exclusion candidates:** CMU `0x40` (`0xc00000bc`) and `0x70`, GPIO `0x44`.

Nothing is guessed — each has the header's answer and the die's side by side. The fixes are mechanical and deliberately a separate change.


w1ne:
Superseded by #1042 — same commit, cherry-picked onto main. #1037 was squash-merged, so this branch's history re-proposed content main already has and GitHub read the whole thing as conflicting. The branch `feat/mg26-silicon-capture` is left in place; nothing is rewritten.

PR #1045, opened 2026-08-21
refactor(bus): InterruptFabric seam — eleven chip interrupt fields leave SystemBus

w1ne:
Closes ledger row 6.6.

## The measurement

Re-derived on `origin/main` (e0c77812c) before touching anything, and again after.

| | before | after |
|---|---|---|
| `SystemBus` `pub` fields | **40** | **38** |
| …of those named for a chip | **2** (`esp32c3_irq_routing`, `esp32s3_irq_routing`) | **0** |
| `SystemBus` fields, all visibilities | **77** | **67** |
| `esp32*_irq_routing` call sites (core + cli) | **69** | **0** under that name |
| files containing `InterruptFabric` | **0** | 6 |
| hand-written `SystemBus { .. }` literals | 8, each listing all 11 fields | 8, each listing 1 |

Counting commands, so the numbers are reproducible:

```
awk '/^pub struct SystemBus \{/,/^\}/' crates/core/src/bus/mod.rs \
  | grep -cE '^    pub [a-z0-9_]+ *:'                     # 40 -> 38
awk '/^pub struct SystemBus \{/,/^\}/' crates/core/src/bus/mod.rs \
  | grep -cE '^    (pub(\(crate\))? )?[a-z0-9_]+ *:'      # 77 -> 67
grep -rn --include='*.rs' -E 'esp32(c3|s3)_irq_routing' crates/ | wc -l   # 69 -> 0
```

(Positive control for the zero: the same `grep -rl` for `SystemBus` still hits
362 files, so the pattern machinery is live and the 0 is a real absence, not a
quoting mistake.)

The row's framing — "two pub fields named for a chip" — undersells it. Eleven
fields on the shared bus were interrupt-fabric state; the three `pub` ones were
only the part a reader of the type signature could see:

```
esp32c3_irq_routing  riscv_irq_lines  esp32c3_system_idx
esp32c3_interrupt_core0_idx  esp32c3_irq_cache  esp32c3_asserted_sources
esp32c3_sched_asserted_sources  esp32s3_irq_routing  esp32s3_intmatrix_idx
esp32s3_asserted_sources  esp32s3_sched_asserted_sources
```

That is also why 69 is not the real blast radius: every reference to all eleven
is ~250, and the eight hand-written `SystemBus { .. }` literals each had to list
all eleven — **80 lines of the diff are pure boilerplate deletion** (88 literal
field lines become 8).

## What moved

One field, `pub irq_fabric: InterruptFabric`, in a new
`crates/core/src/bus/interrupt_fabric.rs`:

| was | is |
|---|---|
| `esp32c3_irq_routing` | `irq_fabric.esp32c3.routing` (pub) |
| `riscv_irq_lines` | `irq_fabric.esp32c3.irq_lines` (pub) |
| `esp32c3_system_idx` | `irq_fabric.esp32c3.system_idx` |
| `esp32c3_interrupt_core0_idx` | `irq_fabric.esp32c3.interrupt_core0_idx` |
| `esp32c3_irq_cache` | `irq_fabric.esp32c3.intc` |
| `esp32c3_asserted_sources` | `irq_fabric.esp32c3.walk_sources` |
| `esp32c3_sched_asserted_sources` | `irq_fabric.esp32c3.sched_sources` |
| `esp32s3_irq_routing` | `irq_fabric.esp32s3.routing` (pub) |
| `esp32s3_intmatrix_idx` | `irq_fabric.esp32s3.intmatrix_idx` |
| `esp32s3_asserted_sources` | `irq_fabric.esp32s3.walk_sources` |
| `esp32s3_sched_asserted_sources` | `irq_fabric.esp32s3.sched_sources` |

`Esp32c3IrqCache` moved with them as `Esp32c3IntcCache`. Field-for-field, no
behaviour change.

Only the two flags the row named stayed `pub` (plus `irq_lines`, which
out-of-crate integration tests and the RISC-V core read). The other eight are
`pub(crate)`, so the crate boundary now tells the truth about who may touch
interrupt-matrix internals.

## What the seam adds beyond a namespace

Two chip-neutral predicates, for the SHARED code that named both SoCs to answer
a question that is about neither:

- **`matrix_owns_cpu_irqs()`** — the classic-ESP32 DPORT aggregation asks this
  before rebuilding `pending_cpu_irqs`, instead of
  `esp32s3_irq_routing || esp32c3_irq_routing`. Three fabrics write that one
  bitmap; exactly one may own it.
- **`per_cycle_aggregation_free()`** — the walk-free per-cycle tick policy
  (`per_cycle_tick_is_trivial`) asks this instead of spelling out both fabrics'
  aggregation conditions inline.

Genuinely chip-specific gates still say which chip. The C3's opt-out of the
Nordic GPIO/GPIOTE service pass is a fact about the C3 — its board inputs write
the C3 GPIO model directly, and ROM boot needs lean ticks — so giving that a
neutral name would be a euphemism, not an abstraction. It now reads through the
fabric (`irq_fabric.esp32c3.routing`) and stops there.

## New gate, and what the negative control actually proved

`bus::tick::classic_dport_defers_to_a_matrix_fabric` — three cases on a
hand-built bus carrying a classic DPORT with matrix source 34 bound to PRO_CPU
slot 9 (TRM §7: PRO map base `0x104`, source `s` at `base + s*4`):

| case | assertion |
|---|---|
| no matrix fabric | DPORT routes: `pending_cpu_irqs == [1 << 9, 0]` |
| `esp32c3.routing = true` | DPORT defers: `pending_cpu_irqs == [0, 0]` |
| `esp32s3.routing = true` | DPORT defers: `pending_cpu_irqs == [0, 0]` |

It exists because nothing held `matrix_owns_cpu_irqs()` to covering *both*
fabrics: no in-tree test exercises a classic DPORT on a bus with S3 routing on,
so a version reading only the C3 flag would have passed everything on main.

**Negative control, run for real, not asserted:**

1. All three green: `cargo test -p labwired-core --lib bus::tick::classic_dport`
   → `3 passed; 0 failed`, exit **0**.
2. Broke exactly one thing — narrowed the predicate body in
   `interrupt_fabric.rs` from `self.esp32c3.routing || self.esp32s3.routing`
   to `self.esp32c3.routing`, i.e. dropped the S3 term.
3. Re-ran: exit **101**, `2 passed; 1 failed`. The failing test was
   `an_active_s3_fabric_takes_ownership_of_pending_cpu_irqs` — its
   `pending_cpu_irqs == [0, 0]` assertion. The other two (`dport_routes_…` and
   `an_active_c3_fabric_…`) stayed **green**, which is the point: the red is
   attributable to the S3 term specifically, not to anything else on the tick
   path.
4. Restored the predicate byte-for-byte (`git diff` on the file: empty) and
   re-ran: back to `3 passed; 0 failed`, exit **0**.

## What I deliberately did NOT do

**No exclusive enum.** `InterruptFabric` holds the two fabrics side by side, not
as `enum { None, Esp32c3, Esp32s3 }`. No SoC has both, and an enum would be the
stronger type — but the two flags are set by *different disciplines*:
`esp32s3.routing` is DERIVED in `rebuild_peripheral_ranges` from the presence of
the intmatrix model, `esp32c3.routing` is ASSERTED by the C3 ROM-boot path, and
the existing `c3_and_s3_interrupt_routing_caches_are_separate` explicitly pins
that setting one never implies or clears the other. Making them exclusive
changes what a bus carrying both is allowed to do — a claim about behaviour, not
a rename — and nothing here measures it as safe. Left as a follow-up that should
carry its own evidence.

**No unified fabric trait.** The two do not share an output: the C3 produces a
RISC-V line mask the core ORs into `mip`; the S3 produces a per-core CPU-slot
bitmap plus an `INTR_STATUS` mirror esp-hal reads back to discover which source
fired. What they *do* share — polling the live level of every scheduler-driven
peripheral — was already one function (`poll_scheduler_matrix_sources`) before
this PR. Flattening the outputs would erase a real difference between two
silicon designs.

**No C3 flag derivation.** Symmetry with the S3 flag is tempting
(`routing = interrupt_core0_idx.is_some()`) and it is a fidelity change: a bus
can carry a declarative `interrupt_core0` peripheral without the ROM-boot wiring
that makes matrix routing correct, and deriving the flag would silently switch
such a bus onto the matrix path. The asymmetry is now *documented on both
fields* instead of being an accident of two authors.

**Left alone — the four `esp32c3_pms*` fields** (`esp32c3_sensitive_idx`,
`esp32c3_pms`, `pms_write_bypass`, `esp32c3_pms_armed`). Permission-control
state, not interrupt fabric, and none of them is `pub`, so they are outside row
6.6 and outside the pub-surface complaint. A PMS seam is its own row.

**Left alone — `crates/core/src/peripherals/esp32*/i2c.rs`.** Confirmed: those
three files are **untouched** by this branch. PR #1043 (`EspI2cCore` extraction)
was rewriting them; this change never needed them.

```
$ git diff --name-only origin/main...refactor/interrupt-fabric | grep i2c
(no output)
```

**Left alone — three stale prose references to `riscv_irq_lines` in
`validation/manifest.yaml` comments.** Editing a gate file for a comment tidy is
review surface for zero functional gain. Noted here rather than done silently.

> **Correction to an earlier draft of this body.** I originally wrote that no
> `crates/core/src/bus/*.rs` path is watched and therefore *this refactor moves
> no board's drift digest*. The first half is true; the conclusion was wrong,
> and `pr-gate` caught it. Three digests move — `esp32c3` via
> `crates/core/src/cpu/riscv.rs`, `esp32s3` via the **directory** entry
> `crates/core/src/peripherals/esp32s3`, and `ci-fixture-riscv` via `riscv.rs`
> too. Both changes are test-only field paths (two assignments in a
> `#[cfg(test)]` fn; one assertion), so `61f9e1db1` records a `drift_ack` with
> that reasoning written into the manifest rather than a silent bump, and it
> does not discharge the re-captures either board already owed.
> `generate_validation_status.py --check` → exit 0. Detail in the PR comments.

## Silicon fidelity

No register offset, interrupt number, source ID, or reset value changed. The
diff is field paths plus one predicate per shared policy site. The C3 still
routes 128 matrix sources through `INTERRUPT_CORE0` MAP (`0x600C_2000 + s*4`) /
`CPU_INT_ENABLE` (0x104) / `CPU_INT_PRI_n` (0x114 + n*4) / `CPU_INT_THRESH`
(0x194), with the SYSTEM `FROM_CPU` doorbells entering as sources 50..53. The S3
still routes the same source IDs through per-core Xtensa MAP tables into
`pending_cpu_irqs` plus the `PRO_INTR_STATUS_REG_n` mirror. Those outputs differ
because the chips differ; the seam holds the difference rather than erasing it.

The two neutral predicates are pure boolean rewrites of the expressions they
replaced (`a || b` reordered, and a conjunction lifted verbatim), so they cannot
change a result.

## Verification status

Green on the merged tree (`ea67639e4`, which includes `origin/main` after #1042
landed under this branch), exit codes captured directly (no pipe):

| check | exit |
|---|---|
| `cargo check --workspace --all-targets` | **0** |
| `cargo clippy --workspace --all-targets` | **0** — no warning in any file this PR touches |
| `cargo fmt --all --check` | **0** |
| `cargo test -p labwired-core --lib bus::tick::classic_dport` | **0** (3 passed) |

All with `PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1` (crates/python pins pyo3 0.20,
host Python is 3.14).

### ⚠️ OWED: the by-name `cargo test --workspace` comparison

**Not yet done. Do not read the table above as a full-suite pass.**

The full-workspace run is `cargo test --workspace --no-fail-fast` on this branch
AND on the same base commit pristine, with the failing sets compared BY NAME.
Two attempts have not completed:

1. First attempt ran fail-fast and aborted at
   `labwired-cli --test demo_blinky` (`BASELINE_EXIT=101`) — an environmental
   red, `Firmware not found at target/thumbv7m-none-eabi/release/demo-blinky`,
   nothing to do with this change. That is why the real run needs
   `--no-fail-fast`.
2. Second attempt (both runs, `--no-fail-fast`) was killed by the host part-way
   through the branch leg — 7 test binaries in, no pristine leg started.

The machine is running several other cargo builds concurrently, which is why
these are taking hours. The comparison is being re-run now and will be posted as
a PR comment. **`--auto --squash` is set, so CI is the gate that actually
decides this merges** — but the local by-name comparison is owed and is not
being claimed.


w1ne:
Full-workspace by-name comparison is **running now** (branch leg started 01:05).

Method, so it can be checked:

```
cd /Volumes/LabWired/worktrees/row-6-6            # branch @ 698e1a776
PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 cargo test --workspace --no-fail-fast > branch.txt 2>&1
echo "BRANCH_EXIT=$?" >> branch.txt              # exit captured directly, not through a pipe

git switch --detach origin/main                    # same base, e0c77812c
PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1 cargo test --workspace --no-fail-fast > pristine.txt 2>&1
echo "PRISTINE_EXIT=$?" >> pristine.txt
```

Both legs in the same worktree at the same base, so the only difference between
them is this branch's two commits. Failing sets are then extracted **by test
name** from each log's `failures:` blocks and diffed three ways — regressions,
fixes, and red-on-both. Counts are reported alongside, but the verdict is the
name diff.

Note on log parsing: these logs carry ANSI escapes and non-UTF8 bytes from
firmware stdout, which silently breaks `grep`/`awk` here (a `grep -c` on the
log returned empty rather than 0). The extractor reads bytes in Python for that
reason.

Known environmental red that will appear in **both** legs, not a regression:
`test_demo_blinky_gpio_toggle` — `Firmware not found at
target/thumbv7m-none-eabi/release/demo-blinky`. It is also why the run needs
`--no-fail-fast`: fail-fast aborts the whole workspace there.

Will post the three lists when both legs finish.

w1ne:
### `pr-gate` was red — fixed in 61f9e1db1, and it was a real catch

The failing step was **"Validation status — regen check + drift gate"**. It was
right to fail: two boards watch content this branch touched.

| board | watched entry | what moved |
|---|---|---|
| `esp32c3` | `crates/core/src/cpu/riscv.rs` | two field **assignments** in a `#[cfg(test)]` fn (`bus.riscv_irq_lines = 1 << 5` / `= 0`) + two comment lines. No non-test line changed. |
| `esp32s3` | the **directory** `crates/core/src/peripherals/esp32s3` | one file, `crosscore_ipi.rs`; within it one assertion in a `#[cfg(test)]` fn. No production line under `esp32s3/` changed. |

Both digests therefore moved on **test-only field paths**. No register offset,
reset value, source ID or interrupt number changed, so the C3's 2026-08-09
87-register reset capture and the S3's 9-register reset-state anchor both still
hold. Recorded as a `drift_ack` with that reasoning written into the manifest —
not a silent digest bump — and it explicitly does **not** discharge the
re-captures each board already owed.

```
esp32c3  4f22db3b82bcf418c24bae16f655e13745970e42f080eb4610eb08c82176ffba
esp32s3  b0cce7fc436b9a1e1b6bb8efc7c6590cb869be690134ac5e61df9d906a9825c6
python3 scripts/generate_validation_status.py --check  -> exit 0
```

`ci-fixture-riscv`'s digest moved too (it also watches `riscv.rs`) but it has no
silicon capture, so it carries no ack — just the regenerated digest column.

**A correction to my own earlier claim.** The PR body says no `crates/core/src/bus/*.rs`
path appears in any board's `models:` list. That part is still true — but I had
also concluded from it that this refactor moves no board's digest, and that was
wrong. It moves three, via `cpu/riscv.rs` and the `peripherals/esp32s3`
**directory** entry. The grep I based it on was a false negative on a log
carrying non-UTF8 bytes; the gate found what I missed. That is the gate working.

Because the tree changed, the workspace comparison legs were stopped and
restarted against the final commit (61f9e1db1). Still owed; will post.

w1ne:
Merged as `f1a0f0dbe` — all checks green, including `pr-gate` and all three `pr-workspace-tests` shards.

The local by-name comparison is **still running** and I am still going to post it. It
changes role now: it is no longer a pre-merge gate, it is post-merge verification of
main. If it turns up a regression, that is a red main and I will act on it rather than
quietly drop it.

One trap worth recording, because it would have made the comparison vacuous: the
pristine leg resolves `origin/main` at switch time, and `origin/main` now points at
**my own squash commit**. A `git fetch` by me — or by any of the several agents sharing
this checkout — before that leg starts would have baselined my change against itself and
produced a confident, meaningless "no regressions". The baseline is therefore pinned to
a tag at `4a2b9fd87` (the pre-merge main this branch merged in), and the leg's actual
HEAD is checked against it before the diff is trusted.

w1ne:
## Branch leg done. And the baseline trap I warned about actually fired.

**The trap fired.** The pristine leg auto-started and its own log line said it was at
**`f1a0f0dbe` — my own squash commit**. Another agent sharing this checkout had fetched,
moving `origin/main` onto the merge of this very PR. That run would have compared my
change against itself and returned a confident, meaningless "no regressions". Killed it
mid-flight, discarded its log, and relaunched pinned to the literal SHA `4a2b9fd87`
(the pre-merge main this branch merged in) rather than to the moving `origin/main` ref.
The relaunched leg prints its resolved HEAD next to the wanted SHA so the pin is
checkable, not assumed:

```
=== PRISTINE run at 4a2b9fd87 (want 4a2b9fd87) started ...
```

### Branch leg: `BRANCH_EXIT=101`, 5 failing tests by name

Run at `ea67639e4` (this branch with `origin/main` merged in),
`cargo test --workspace --no-fail-fast`, `PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1`,
exit captured directly.

| test | why it failed |
|---|---|
| `test_demo_blinky_gpio_toggle` | `Firmware not found at target/thumbv7m-none-eabi/release/demo-blinky` |
| `test_dap_e2e_launch_and_uart` | `fixture missing — build firmware-ci-fixture first` (its own message points at the `core-ci.yml` step that builds it) |
| `real_scanner_and_ecu_firmware_complete_the_obd2_workflow` | `build the real scanner ELF: NotFound` |
| `test_strict_board_onboarding` | `[FAIL] Firmware build failed for stm32l073` — cross-compile, every other board PASSes |
| `every_cpu_core_is_covered_by_this_file` | `cores implementing Cpu` = `[Avr, CortexM, RiscV, XtensaLx7]` vs exercised `[CortexM, RiscV, XtensaLx7]` — an `Avr` core exists that was never added to `cores()` |

Four are missing prebuilt firmware artifacts that CI builds in a separate step; one is a
pre-existing coverage gap for a CPU core this PR did not add. None of the five touches
interrupt routing, and none is in a file this PR modified.

**That reasoning is not the proof, though, and I am not offering it as one.** The claim
"not my regression" is only established by the same test being red at `4a2b9fd87`. That
leg is running now; I will post the three-way name diff — regressions / fixed /
red-on-both — when it lands. Until then these five are *unattributed*, not *excused*.

PR #1046, opened 2026-08-21
feat(arduino-matrix): put BRD2709A on the board, and let a lane bring its own compiler

w1ne:
BRD2709A was the only LabWired-supported Arduino target with no matrix column,
and the reason was structural rather than an oversight: `run_matrix.py`
compiles with PlatformIO, and Silicon Labs never published an xG26 Arduino
core, so LabWired's is hand-written and lives in the labwired monorepo. Its
compile flags — the hard/soft float split, the 0x08008000 app slot — were
derived on real silicon and get exactly one home. Copying them into this repo
would give them two, and two is how they drift.

So a board may now declare `external_compile:` instead of `pio:`, and the lane
that owns the toolchain supplies the driver through `$LW_EXTERNAL_COMPILE`.
⚠️ A missing driver is `toolchain_missing`, which the gate COUNTS — not a
skip. `crates/core/tests/arduino_matrix_lanes.rs` is what stops the other
half of the hole: a board in boards.yaml must either be in this repo's job
matrix or name the workflow elsewhere that runs it, and a board that cannot
compile here must not be in this repo's job matrix at all. Proven able to
fail — deleting the `lane:` line reds two of its three tests.

## What the first run found

Six of the nine kits pass: L0, L1, L2, L3 (I²C INA219 through Wire on I2C0 at
0x4B000000), L5 (analogRead), L7 (micros). Getting there took three fixes to
the Arduino core, and all three are the point of running STOCK sketches:

  * `'F' was not declared in this scope` — the core had no `F()` macro, so two
    kits failed to compile pointing at the SKETCH's own line.
  * `'SS' was not declared in this scope` — the variant declared no SPI pin
    names at all.
  * `'snprintf' was not declared` — the lane linked `-nostdlib`, so four of the
    nine kits could not link, with the error naming newlib's sbrk.c.

None of that was visible before, because everything that had ever compiled for
this board was written by someone who knew what the core supported. A core that
cannot build an ordinary Arduino sketch is not an Arduino core. (Those three
fixes are in the monorepo, alongside the core they fix.)

## What it found that is NOT fixed

L4 and L6 compile, boot, print their BOOT marker and then hang in `delay()`,
spinning on TIMER0_CNT while CNT does not advance — measured stuck at
`delay+0x14` with r5=1000 and CNT-start = 0 after **30,000,000** steps. The
first `delay(1)` completes; the stall begins only after `SPI.begin()` or
`analogWrite()`. Hand-written sketches doing the same calls in the same order
do not reproduce it, including a silent 30x `delay(1)` loop with PWM running,
so the trigger is narrower than "a second peripheral is active".

They are skipped WITH THAT SENTENCE rather than dropped from the board, so the
six real cells gate and the two open ones stay visible. Written up in
validation/arduino-matrix/PROBLEMS.md.

---

_Stacked on #1042 in spirit but branched from it — merge #1042 first if both are open. The companion monorepo PR (compile driver + workflow + the three Arduino-core fixes) is w1ne/labwired#1814._


w1ne:
Superseded by #1047 — same commit on today's main. #1042 was squash-merged while this was open, so the branch re-proposed content main already has.

PR #1067, opened 2026-08-25
Bus observers see peripheral MMIO stores; machine observers reach the bus

CrispStrobe:
## What

Two gaps together made a `--vcd` trace record `pc` and nothing else:

1. **`SystemBus::write_u32`/`write_u16`'s peripheral branches never called `on_memory_write`**, so 32-bit MMIO stores — GPIO BSRR, USART TDR, exactly the traffic a firmware trace exists for — were invisible to observers. They now notify per byte, little-endian, the same shape the RAM fallback already produces via `write_u8`. `old` is passed as 0 on purpose: reading a peripheral register to report its previous value can have side effects (RDR, status-clear-on-read).

2. **`Machine.observers` and `SystemBus.observers` are separate lists** and nothing mirrored one into the other, so even RAM-path writes never reached a CLI-registered observer (the CLI pushes the `VcdObserver` into `machine.observers` only). `Machine::advance` now mirrors them (Arc clones, idempotent per call).

## How it was found

A differential-testing harness runs the same STM32F030 ELF (TIM3 millisecond tick → NVIC IRQ → WFI idle → GPIO blink + USART banner) on LabWired and on an independent Cortex-M0 simulator and compares pin-edge timelines. With the VCD empty of MMIO there was nothing to compare; with this patch, the reassembled BSRR edge timeline agrees with the second simulator edge-for-edge, and the cross-simulator timebase ratio measures 47.999 cycles/µs against the nominal 48 — the same 5 ms blink grid to 0.002% from two independently written models.

## Notes

- Byte-granular events keep the existing `on_memory_write(addr, old, new)` signature untouched; an 8-bit-data VCD consumer sees each byte at `addr..addr+3`.
- Happy to split the two changes, gate the mirroring differently, or add a test in your preferred shape.

Found while using LabWired as a differential oracle for the BrickWright educational simulator (MIT). Thanks for the project — the stm32f0 onboarding config plus `test` mode made a cross-simulator CI gate genuinely easy to stand up.


w1ne:
Thanks for this — both gaps you found are real, and the differential-harness write-up is what made them easy to confirm. The blast radius is actually a bit wider than you framed it: `TraceObserver` implements `on_memory_write` and is registered the same way, so `--trace` has been recording no memory writes either, not just `--vcd`.

I'm closing this in favour of #1068, which keeps your commit and co-authorship and reshapes both fixes. The reason it can't go in as-is:

`write_u16`'s peripheral branch ended up with two notification loops — the 2-byte one and a copy of the 4-byte one from `write_u32`. Shifting a `u16` right by 16 and 24 is an arithmetic overflow, and this workspace sets `overflow-checks = true` in **both** profiles (see the `[profile.release]` comment in the root `Cargo.toml`) with `panic = "abort"`:

```
panic_const_shr_overflow
  at labwired_core::bus::accessors::<impl Bus for SystemBus>::write_u16 (accessors.rs:589)
```

An ordinary `TIMx_ARR` write is enough to hit it, and no observer needs to be attached — the shift is evaluated before the observer loop. `bus_visibility` already drives a `write_u16` at an SPI `CR1`. The same minimal test passes on `main` and aborts on `bw/mmio-write-observers`, so it's a regression rather than something pre-existing, and it's in the original commit rather than the merge.

Two things changed in #1068 beyond the obvious de-duplication:

1. The byte count now comes from `value.to_le_bytes()` rather than a literal at the call site, so the width can't be restated wrong. That's the property that makes this class of bug not recur, and it's why I didn't just delete the stray loop.

2. The `advance()` mirroring compares the two lists by *length* and then assigns over `bus.observers` — that evicts an observer registered on the bus directly (the DAP memory tracker is one), and it only converges once a run has started, so an embedder that writes before advancing still sees nothing. Since `on_memory_write` is emitted only by the bus and step/tick events only by the `Machine`, the lists carry disjoint event kinds and the split is really just a borrow-checker artifact. So the fix moved to registration: `Machine::add_observer` appends to both, `SystemBus::add_observer` covers bus-only observers, and `advance()` goes back to only advancing.

You offered to split the change or add tests in our preferred shape — that's essentially what #1068 does, so no further work needed on your side. Your `old = 0` reasoning was right and is preserved, with your rationale in the rustdoc.

Also worth flagging: no CI checks ran on this PR at all, which is how an aborting change reached a clean-looking mergeable state. Worth a look separately.

---

PR #1095, opened 2026-09-03
event-scheduler migration phases 1+2: 30 duplicated cfg! sites retired, 8 blockers reclassified

w1ne:
Two commits, one CI cycle. Commit 1 is a pure refactor; commit 2 changes behaviour on three boards.

Every number below was re-derived by reading files in this branch, with the gate's own counter or a Python cross-check validated against it — none is quoted from anywhere else.

---

## Phase 1 — retire 30 duplicated `cfg!` sites (`d5b2b771`)

**Zero behaviour change.**

34 types under `crates/core/src` define a private inherent `fn scheduler_mode(&self) -> bool`. Hashing each body (brace-matched, whitespace-normalised) gives 30 sharing one digest:

```
{ cfg!(feature = "event-scheduler") && self.clock.is_some() }
```

Those 30 are now one macro in `crates/core/src/cycle_clock.rs`, invoked as `crate::cycle_clock::scheduler_mode!();` where each copy sat.

### Mechanism chosen: a macro, not a trait default

`scheduler_mode` is a **private inherent** method on 30 unrelated peripheral types. A trait default would have to:

* widen its visibility from private to at least `pub(crate)`;
* add a `clock()`-style accessor impl per type — ~30 x 3 lines, giving back most of what the 30 x 4-line deletion buys;
* add a `use` line to each of the 26 files that call it;
* and put a trait method into name-resolution competition with the four models that keep a hand-written `scheduler_mode` of their own.

A macro expanded inside the existing `impl` keeps the method private, inherent and callable exactly as before: **no call site changes and no name-resolution change anywhere**. It lives in `cycle_clock.rs` because the predicate is precisely "the feature is on AND this instance holds a bus-published `CycleClock`", and it is crate-private (`pub(crate) use scheduler_mode`), invoked by path so no file needs a new import.

### The 4 left alone

Not stale copies — genuinely different bodies:

| site | why it differs |
|---|---|
| `Exti::scheduler_mode` | reads `clock` through an enum match |
| `Fdcan::scheduler_mode` | `&& self.bus_rx.is_none()` — single-node path only |
| `I2c::scheduler_mode` | delegates to the selected variant |
| `nrf52::Timer::scheduler_mode` | `self.clock.is_some()`, **no `cfg!` at all** |

### Doc comments

21 distinct doc blocks sat above the 30 copies; 19 restate the same sentence. That sentence now lives once, on the macro. The two that say something the macro's doc does not — the EFR32 `I2c` "no cycle-paced engine" note and PIO's "every SM freezes at PC 0" note — are kept verbatim at their call sites as `//` comments (a `///` on a macro invocation is a dead doc comment).

### Ratchet

| ceiling | before | after |
|---|---|---|
| `MAX_MODEL_SITES` | **208** | **179** |
| `MAX_HARNESS_SITES` | **75** | **75** (unchanged) |

Measured with the gate's own counter, by temporarily zeroing the ceilings and reading the assertion message:

* base (`77922314`): 140 attributes (9 negated) + 68 `cfg!` = 208, over 77 files with a site.
* after: the same 140 attributes (9 still negated) + 39 `cfg!` = 179, over 64 files.

Net −29: 30 `cfg!` sites removed, 1 added inside the macro. Cross-checked with a Python port of `count_in_source` that reproduces the Rust gate exactly on both trees (208/140/68 and 179/140/39, and harness 75/70/5 both times). The module doc's derived figures were re-derived from the same counter rather than left stale.

---

## Phase 2 — reclassify 8 blockers that were never migration work (`7f38a05b`)

**This changes behaviour.** `derive_walk_deletable()` is `all(uses_scheduler() || !needs_legacy_walk())`, so a board whose last blocker was one of these now derives walk-deletion.

### The nine, re-verified

For each, the `impl Peripheral` block was read in full and its overridden methods enumerated by script (regex covering path-qualified `impl crate::Peripheral for …` too):

| type | site | overrides any tick? | verdict |
|---|---|---|---|
| `AvrGpioPort` | `peripherals/avr_gpio.rs:42` | no | `false` |
| `Esp32I2cAhbFifo` | `peripherals/esp32/i2c.rs:319` | no | `false` |
| `FlashSpiMemStub` | `.../system_stub.rs:148` | no | `false` |
| `TimgStub` | `.../system_stub.rs:388` | no | `false` |
| `RomThunkBank` | `.../rom_thunks.rs:219` | no | `false` |
| `SimCtl` | `peripherals/simctl.rs:142` | no | `false` |
| `OneShotRam` | `.../rom_thunks.rs:2358` (test-local) | no | `false` |
| `RamPeripheral` | `boot/esp32s3.rs:282` (test-local) | no | `false` |
| `CosimPeripheral` | `cosim/mod.rs:81` | `tick` = `default()` + TODO | **left `true`** |

None of the nine overrides `tick`, `tick_elapsed`, `tick_elapsed_forced`, `tick_with_bus`, `needs_bus_tick`, `legacy_tick_active`, `uses_scheduler`, `idle_poll_bus_tick` or `route_ppi_events` — except `CosimPeripheral::tick`. So the walk calls the trait default, whose `PeripheralTickResult` has `irq: false`, no `dma_requests`, no `explicit_irqs`, no `system_exception`, no `dma_signals`, an empty `mmio_writes` and an empty `fired_events`. The "emits nothing from the walk" half of the contract is structural here, not a judgement call.

Per-model reasons are in doc comments at each site (register bank; MMIO-driven calibration state machine; read-only ROM whose thunks dispatch from the CPU's `BREAK` arm, never from a clock; `SimCtl`'s `SCLK` served lazily from the bus-published `CycleClock` on a `&self` read, with the exit code drained by `Machine::advance`, not by the walk).

**`Esp32Uart` was checked and is NOT on the list**, as flagged: its one-line `tick` is `self.tick_elapsed(1)` and it overrides both. It stays a walk-forcer.

### Left `true`: `CosimPeripheral`

Nothing in the workspace constructs it — repo-wide Python scan over 4,890 files, positive control `needs_legacy_walk` = 324 occurrences — so a `false` banks nothing on any bus. And the TODO on its `tick` says the intended body syncs simulation time with an external process, i.e. real walk work. A `false` left standing there is a pre-planted starvation bug for whoever fills it in, which is exactly the failure the contract warns about. The reason is recorded at the site so the next inventory does not re-litigate it.

### Correction to the brief

`system::xtensa::RamPeripheral` — the **production** flat window — already overrode `needs_legacy_walk() -> false`. The `RamPeripheral` that was still `true` is only the test double inside `boot/esp32s3.rs`'s `#[cfg(test)] mod tests`. Its impl header is `impl crate::Peripheral for RamPeripheral`, which is how a name-only search misses it. Also: the brief said 35 `scheduler_mode()` definitions; there are **34** — the 35th grep hit is a test function named `scheduler_mode_write_sync_and_clock_sync_do_not_double_count`.

### Which boards' `derive_walk_deletable()` flips

All 280 system manifests under `configs/systems/**` and `examples/**` were built with `walk_deleted = None` and their forcer sets printed, before and after. Under `--features event-scheduler`, **exactly three** change, and nothing else moves in either direction:

| board | before | after |
|---|---|---|
| `configs/systems/arduino-nano.yaml` | `false`, forcers `["portb"]` | **`true`**, forcers `[]` |
| `configs/systems/simctl-selftest.yaml` | `false`, forcers `["simctl"]` | **`true`**, forcers `[]` |
| `configs/systems/pico-selftest.yaml` | `false`, forcers `["simctl"]` | **`true`**, forcers `[]` |

Without `event-scheduler`, only `simctl-selftest` derives differently — and there the flag is never consulted, because `tick_peripherals_phase1` reads `legacy_walk_disabled` under `#[cfg(feature = "event-scheduler")]`. That build is unchanged.

**Classic ESP32 does not flip.** Its production bus (`configure_xtensa_esp32`, not `from_config` — `from_config` stubs the xtensa models and would report a false roster) goes from 6 walk-forcers to 4:

```
before: ["rom", "uart0", "uart1", "uart2", "i2c0", "i2c0_ahb_fifo"]
after:  ["uart0", "uart1", "uart2", "i2c0"]
```

`rom` is the `RomThunkBank`, `i2c0_ahb_fifo` the `Esp32I2cAhbFifo`. The three `Esp32Uart` instances and `Esp32I2c` still hold the walk, correctly. Because the derivation stays `false`, no classic-ESP32 behaviour changes — the blocker list simply names only real blockers now. ESP32-S3 was already at zero forcers and is unmoved.

### Evidence for the three that flip

* **arduino-nano** — golden ELF run for 500,000 steps on a bus built with `walk_deleted = None`, and again with `walk_deleted = Some(false)` (walk pinned on). Byte-identical: same cycles (500000), same PC (`0x1ca`), same UART bytes (`"nano-ok\r\n"` + the blink dots). Its three peripherals are `spi` and `i2c` (both `uses_scheduler()`, so the walk already returned `default()` for them) plus `portb`, whose tick was already the no-op default — deleting the loop removes no work.
* **simctl-selftest** — `cargo test -p labwired-cli --features labwired-core/event-scheduler --test simctl_selftest_dogfood` passes 2/2, including `the_selftests_assertion_actually_gates`. The same walk-on / walk-derived differential is identical.
* **pico-selftest** — stated plainly: **no firmware fixture for this board exists in-tree**, so this is a derivation result, not a firmware differential. It is `pico.yaml` plus the `simctl` window, and `pico.yaml` is already walk-deleted today, so the flip puts no new peripheral on a walk-free bus.

### Ceilings after Phase 2

Unchanged — `MAX_MODEL_SITES` stays **179**, `MAX_HARNESS_SITES` stays **75**. Adding a `needs_legacy_walk` override removes no conditional-compilation site.

---

## Gates run locally (all green)

* `--lib` under all three feature sets: default (**3308** passed), `event-scheduler` (**3438**), `jit,event-scheduler` (**3485**).
* Every per-chip walk gate under `--features event-scheduler`: `walk_deletion`, `tick_interval_inventory`, `board_batch_width`, `bench_walk_free_kw41z`, `gpio_devices_walk_free`, `kw41z_walk_free_differential`, `systick_`, `efr32mg26_`, `esp32_classic_`, `esp32c3_`, `esp32s3_`, `nrf52_timer_`, `nrf54l15_grtc_`, `stm32_dma_`, `stm32_timer_`, `stm32f103_dma_`, `stm32f401_`, `stm32f407_dma_`, `stm32h563_`, `stm32l073_`, `stm32l476_walk_differential`, `esp32c3_bt_active_walk_pinners`.
* The CI scheduler-only lane list: `esp32c3_bt_event_residency_gate`, `esp32s3_irq_cache_differential`, `esp32s3_smp_yield_latency`, `nrf52_easydma_tick512_fidelity`, `nrf52840_timer_machine_gate`, `rp2040_timer_machine_gate`, `rp2040_i2c_irq_delivery`.
* `avr_nano_machine_run`, `avr_nano_golden_survival`, `esp32_classic_aids_stability`.
* `cargo clippy -p labwired-core --lib` in both feature worlds, `cargo check --lib --tests` in both, `cargo fmt --all --check`.

Known local gap: the ten `esp32c3_walk_differential` cases that are `#[ignore]`d pending `--release --ignored` (they run the real C3 bootloader for ~30M steps) were not run. No ESP32-C3 model is touched by either commit.


w1ne:
## CI status: one blocker, and it needs a human ack

Everything green except `pr-gate`, which fails at **step 9, "Validation status — regen check + drift gate"** (39s, before any of its own Rust steps ran). Passing: `pr-workspace-tests` shards 1/2/3 (the feature-unified lane, so every `required-features = ["event-scheduler"]` target ran), `pr-scheduler-observable`, `browser-layer`, `fmt-autofix`, `release-runner-contract`, `actionlint`, `shellcheck`.

### What fails, and why it is not a defect in the change

Nine silicon-tier boards watch model files this PR edits. `drift_ack_digest` hashes model **content**, so any byte that moves stops the ack covering the tree — by design.

I verified the acks covered `origin/main` exactly (`recorded == digest(base)` for all nine) and stopped covering only after these two commits, so **main was green and this PR caused all nine**. Every one is attributable to a file this PR actually touched — none drifted for an unrelated reason:

| board | ack | moved by |
|---|---|---|
| `stm32f103` | 2026-09-02 | c1: `peripherals/{adc,dma,i2c,timer}.rs` |
| `stm32f407` | 2026-09-02 | c1: `peripherals/{dma,i2c,timer}.rs` |
| `nucleo-l073rz` | 2026-09-02 | c1: `peripherals/{dma,i2c,timer}.rs` |
| `nucleo-l476rg` | 2026-09-02 | c1: `peripherals/{adc,dma,i2c,rtc_v3,timer}.rs` |
| `stm32h563` | 2026-09-02 | c1: `peripherals/{adc,gpdma,i2c,rtc_v3,timer}.rs` |
| `nrf52840` | 2026-08-22 | c1: `peripherals/nrf52/rtc.rs` (watched via `peripherals/nrf52`) |
| `seeed-xiao-nrf52840-sense` | 2026-08-22 | c1: `peripherals/nrf52/rtc.rs` (watched via `peripherals/nrf52`) |
| `esp32c3` | 2026-08-22 | c1: `peripherals/esp32c3/{bt,ledc,rtc_timer,wifi_mac}.rs`; c2: `peripherals/esp_xtensa_common/{rom_thunks,system_stub}.rs` |
| `esp32s3` | 2026-08-23 | c2: `peripherals/esp_xtensa_common/{rom_thunks,system_stub}.rs` |

In commit 1 those files changed only by the macro fold — the emitted method is token-for-token what it replaced. In commit 2 the two `esp_xtensa_common` files gained a `needs_legacy_walk() -> false` on `RomThunkBank`/`FlashSpiMemStub`/`TimgStub`, and neither the ESP32-C3 nor the ESP32-S3 bus changes its derivation as a result (S3 was already at zero walk-forcers; C3 instantiates none of the three).

No ack has expired: the earliest lapse is `nrf52840` / `esp32c3` on 2026-09-21.

### The remedy, which I did not apply

`validation/manifest.yaml` says acking is a human assertion, and the generator's own docstring says so too ("stamping an ack is `--write-ack-digests`, and acking is a human act"). I was also denied permission to run it. So I have left the manifest untouched rather than hand-editing the digests around the tool.

The move that matches precedent (`a4936b2f`, "re-stamp the six digests this trait move touched") is a **digest re-stamp with the ack DATES unchanged**, so no promise is extended and each board still lapses on its original schedule:

```
python3 scripts/generate_validation_status.py --write-ack-digests
./scripts/regen-all.sh          # regenerates docs/boards/VALIDATION_STATUS.md
git status                      # expect: manifest.yaml + VALIDATION_STATUS.md only
python3 scripts/generate_validation_status.py --check --drift   # expect exit 0
```

Two things worth checking when you run it:

1. `docs/boards/VALIDATION_STATUS.md` **must** be regenerated and committed as well. Its digest column carries the current model digest independent of the ack, so `pr-gate` step 13 ("Generated artifacts — regenerate and assert no drift") fails without it. I ran `./scripts/regen-all.sh` locally to confirm that doc is the only artifact the change makes stale — nothing else in the tree moved.
2. Six boards carry `drift_ack_digest` values already stale on `main` and are **not** drifted (`rp2040`, `rp2350`, `nrf5340`, `stm32h735`, `stm32f411ceu6`, `mkw41z4`). The tool should now skip them — the "⚠️ ONLY DRIFTED BOARDS ARE STAMPED" restriction landed after the sweep that caused the manual reverts in `a4936b2f`. Worth confirming it stamps exactly nine.

PR #1169, opened 2026-09-17
refactor(riscv): split step into per-class exec modules

w1ne:
## Summary
- Move `RiscV::step`'s instruction match, verbatim per arm, out of `crates/core/src/cpu/riscv.rs` into per-class `exec_*` methods under `crates/core/src/cpu/riscv/exec/{alu,branch,load_store,muldiv,atomic,system,misc}.rs`, mirroring the Cortex-M split landed in #1162.
- `step` keeps fetch/decode, interrupt/trap entry, cycle accounting and the pc-commit tail, and now dispatches by instruction class.
- No behaviour change, no public API change. All 83 `Instruction` variants are accounted for across the seven modules.

## Mechanical deviations required by the move (documented in `system.rs`)
- Several system-class arms (`Ecall`/`Ebreak`, `Mret`, `Csrr*`) originally did `return Ok(());` directly out of `step` to skip the timer-update/interrupt-check/pc-commit tail. A function boundary can't reach through to the caller's `return`, so `exec_system` returns `Ok(true)` in exactly those cases, and `step` does `if self.exec_system(..)? { return Ok(()); }`.
- `next_pc` becomes a `&mut u32` parameter, deref'd as `*next_pc` in the branch arms that write it.
- Every other arm body is byte-identical to before the move.

## Drift gate
`cpu/riscv.rs`'s digest moved (whole-file hashing). Added `crates/core/src/cpu/riscv` to the `models` list for the `esp32c3` and `ci-fixture-riscv` boards, and stamped a dated `drift_ack` (with a fresh `drift_ack_digest`) on `esp32c3` documenting that no model behaviour changed — this is an ACK, not a capture.

## Verification
- `cargo test -p labwired-core --lib`: 3539 passed, 0 failed (contract scanners incl. `tests::` all green)
- `cargo test -p labwired-core --test riscv_jit_{alu,branch,mem,combined}_lockstep --test riscv_jit_lockstep --features jit,jit-framework,event-scheduler`: all pass
- `cargo test -p labwired-core --test esp32c3_walk_differential --test session_builder_riscv --test e2e_esp32c3_snapshot_resume --test diag_esp32c3_boot --features event-scheduler`: all non-ignored pass (ignored ones need ROM/firmware fixtures not in the tree)
- `cargo clippy -p labwired-core --lib --no-deps -- -D warnings`: clean
- `rustup run 1.95.0 cargo fmt --all -- --check`: clean
- `python3 scripts/generate_validation_status.py --check --drift`: exit 0

## Test plan
- [x] CI green on this branch


w1ne:
Closing on measurement, not on correctness (all tests and the JIT lockstep suites pass).

`scripts/perf/board_perf.py --boards esp32c3 --modes batch`, callgrind Ir/step, same machine, same toolchain:

| tree | esp32c3 batch |
|---|---|
| current main (after #1173) | 258.9 |
| this branch, `exec_*` with `#[inline(always)]` | 272.2 (+5.1%) |

Same shape and same cause as the Cortex-M split reverted in #1173: inlining removes the call, but the second variant match inside each class function and `next_pc` passed by `&mut` stay on the hot path. The gate tolerance is 3%.

The AVR split (#1170) is different: `step_inner` is a sequential opcode-mask chain, so inlining the pieces reproduces the original codegen (+0.14%). A future attempt for RISC-V or Cortex-M needs a shape that keeps ONE match in `step` (per-instruction `#[inline(always)]` bodies, for example) and a board_perf table in the PR.

